PR #7, opened 2016-01-12
GPIO OTYPER is only 16 bits

trainman419:
The upper 16 bits of the GPIO OTYPER are reserved; only the lower 16
bits are used. Mark the upper 16 bits as reserved.


pat-rogers:
Well spotted.  Merged.

PR #8, opened 2016-01-22
Fixed error in calculation of the USART BRR register value.

nesos:
The calculation of the BRR value to set up the baud rate of an USART interface uses a wrong factor for the `Int_Divider` calculation. This always results in twice the baud rate as requested by the `To` parameter. 
This fix is consistent with the formula used in the official STM32F4 HAL by ST.

### DIFF
--- a/ARM/STMicro/STM32/drivers/stm32f4-usarts.adb
+++ b/ARM/STMicro/STM32/drivers/stm32f4-usarts.adb
@@ -131,7 +131,7 @@ package body STM32F4.USARTs is
 
    procedure Set_Baud_Rate (This : in out USART; To : Baud_Rates) is
       Clock        : constant Word := APB_Clock (This);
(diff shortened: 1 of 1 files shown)


Fabien-Chouteau:
Actually I think that the value depends on the status of OVER8 in CR1. Can you confirm?

Thanks!

pat-rogers:
On 1/24/2016 11:37 AM, Fabien-Chouteau wrote:
> Actually I think that the value depends on the status of OVER8 in CR1. Can you confirm?

Yes, that right.  I dug up the standard peripheral library's version and 
it shows that difference clearly, unlike that of the HAL.  I'll put that 
handling in place in the Ada version.

Good catch Thibaut. :-)

Best regards,

Pat

pat-rogers:
I've pushed a fix along with a new routine to control the oversampling mode.

PR #10, opened 2016-02-10
Change the project architecture to emphasis both on device and board

Fabien-Chouteau:
The current approach only defines boards (inside a single project). This
patches adds devices projects that will help custom board definition
(e.g. Crazyflie or Embrick).

A project for Cortex-M packages is also added.

The component directory is moved at the root of the repository, since
component could be share across architecture, in particular those
connected on serial buses (SPI, I2C, UART, etc.).

Signed-off-by: Fabien Chouteau <chouteau@adacore.com>


Fabien-Chouteau:
I like the fact that it's easier to define a board outside of this repository. Also there are less scenario variables to mess with.

Tell me what you think.

Matthias-R:
I like the idea of sorting/separating mcu specific, board specific and mcu/board independent parts, too. It should make modifications for other (own) boards easier.
Nice work!

Regards,

Matthias

PR #14, opened 2016-04-12
STM32F4-GPIO : Modify procedure and function to allow non-variable GPIO_Point parameters

VLambret:
Hi,

I noticed some functions where parameters are in out in the signature when the body only read the parameters value. As I was passing non variable parameters it causes compilation errors fixed by this commit.

I take a look at your svd branch where the procedure and function signatures are modified in the same way.

I'm new in the Ada so I may have missed something. I would be grateful for any remark explaining how to improve things.

Best regards,

### DIFF
--- a/ARM/STMicro/STM32/drivers/stm32f4-gpio.adb
+++ b/ARM/STMicro/STM32/drivers/stm32f4-gpio.adb
@@ -126,7 +126,7 @@ package body STM32F4.GPIO is
    -- Set --
    ---------
 
-   procedure Set (This : in out GPIO_Point) is
+   procedure Set (This : GPIO_Point) is
    begin
       This.Port.BSRR_Set := GPIO_Pin'Enum_Rep (This.Pin);
    end Set;
@@ -157,7 +157,7 @@ package body STM32F4.GPIO is
    -- Clear --
    -----------
 
-   procedure Clear (This : in out GPIO_Point) is
+   procedure Clear (This : GPIO_Point) is
    begin
       This.Port.BSRR_Reset := GPIO_Pin'Enum_Rep (This.Pin);
    end Clear;
@@ -196,7 +196,7 @@ package body STM32F4.GPIO is
    -- Toggle --
    ------------
 
-   procedure Toggle (This : in out GPIO_Point) is
+   procedure Toggle (This : GPIO_Point) is
    begin
       This.Port.ODR := This.Port.ODR xor GPIO_Pin'Enum_Rep (This.Pin);
    end Toggle;
--- a/ARM/STMicro/STM32/drivers/stm32f4-gpio.ads
+++ b/ARM/STMicro/STM32/drivers/stm32f4-gpio.ads
@@ -127,7 +127,7 @@ package STM32F4.GPIO is
    --  For the given GPIO port, sets all of the output data register bits
    --  specified by Pins to one. Other pins are unaffected.
 
-   procedure Set (This : in out GPIO_Point) with Inline;
+   procedure Set (This : GPIO_Point) with Inline;
    --  For This.Port.all, sets the output data register bit specified by
    --  This.Pin to one. Other pins are unaffected.
 
@@ -140,7 +140,7 @@ package STM32F4.GPIO is
    --  For the given GPIO port, sets of all of the output data register bits
    --  specified by Pins to zero. Other pins are unaffected.
 
-   procedure Clear (This : in out GPIO_Point) with Inline;
+   procedure Clear (This : GPIO_Point) with Inline;
    --  For This.Port.all, sets the output data register bit specified by
    --  This.Pin to zero. Other pins are unaffected.
 
@@ -153,7 +153,7 @@ package STM32F4.GPIO is
    --  specified by Pins (ones become zeros and vice versa). Other pins are
    --  unaffected.
 
-   procedure Toggle (This : in out GPIO_Point) with Inline;
+   procedure Toggle (This : GPIO_Point) with Inline;
    --  For This.Port.all, negates the output data register bit specified by
    --  This.Pin (one becomes zero and vice versa). Other pins are unaffected.
 


Fabien-Chouteau:
I think this is on purpose because those procedures are modifying the state of the peripheral.

VLambret:
Thanks for your quick reply,

Is this semantic layer needed as it looks quite logic that Set, Clear and Toggle procedure may modify the pin state ?

Maybe I'm thinking too low-level here or lacks of Ada practice, I just read the out keyword as the modification of the parameter, not of the concept it handles.

Is there a project guideline available somewhere for new contributors ? I would prefer respect the project choices for future contributions.

Fabien-Chouteau:
I would not say it's needed because there's nothing in the language that requires us to do that (otherwise the compiler would complain). But if we can use language features to give more sense to the interface, that sounds good to me.

That's my opinion, I will let the others comment :)

We are working on the guidelines, our objective is to have a clear definition and stable interface by the end of May.

Thanks for your contribution!

pat-rogers:
It was done that way purely to suggest that it is modifying the state of the port, as Fabien says, but really it could be done either way because the component is an access value.  For the sake of using constant GPIO_Point values I agree it would be convenient to use mode "in" here.  However, the style for mode "in" formals is to take it as the default, rather than explicitly indicating it.  Could you change your proposal accordingly?  That is, just remove the mode indicator altogether.  We can then merge it.

VLambret:
Of course, I just pushed a modified commit with your proposals.

pat-rogers:
Done.  Thanks for this.

elgor:
Right now, the signature is again with `in out` indicator. Is there a reason behind it or was the commit overwritten by the svd/master merge? I also think it would be nice to use the interface with constant GPIO_Points.

Fabien-Chouteau:
I think I'm the one to blame. STM32.GPIO.GPIO_Point is now implementing the HAL.GPIO.GPIO_Point interface and I defined this interface with "in out" for the clear and set procedures.

I kinda like the possibility to control whether or not a piece of code can change the state of a GPIO_Point.

What is the reason for you to have the GPIO_Points constant?

pat-rogers:
On 5/12/2016 5:02 AM, Fabien wrote:
> I think I'm the one to blame. STM32.GPIO.GPIO_Point is now implementing the HAL.GPIO.GPIO_Point interface and I defined this interface with "in out" for the clear and set procedures.
>
> I kinda like the possibility to control whether or not a piece of code can change the state of a GPIO_Point.

It isn't an easy decision, and I had it the same way as your interface 
originally.

The indirection via the access component makes it an odd case because we 
aren't changing the point, we're changing the state of the GPIO port it 
designates. So mode in-out looks like we are doing something we are not.

> What is the reason for you to have the GPIO_Points constant?

We want to say this, for example:

    Pattern : constant array (Index) of User_LED := (Orange, ..., Green);

It is poor design for it not to be a constant, but we can't because the 
routines set/clear the LEDs use mode in-out.

So I changed the modes in the original, since it allows use like the 
above and doesn't lose anything since we aren't really changing the 
actual parameter anyway.

pat-rogers:
On 5/13/2016 12:44 AM, Pat Rogers wrote:
> On 5/12/2016 5:02 AM, Fabien wrote:
>> I think I'm the one to blame. STM32.GPIO.GPIO_Point is now implementing the HAL.GPIO.GPIO_Point interface and I defined this interface with "in out" for the clear and set procedures.
>>
>> I kinda like the possibility to control whether or not a piece of code can change the state of a GPIO_Point.
>
> It isn't an easy decision, and I had it the same way as your interface
> originally.
>
> The indirection via the access component makes it an odd case because we
> aren't changing the point, we're changing the state of the GPIO port it
> designates. So mode in-out looks like we are doing something we are not.
>
>> What is the reason for you to have the GPIO_Points constant?
>
> We want to say this, for example:
>
>     Pattern : constant array (Index) of User_LED := (Orange, ..., Green);
>
> It is poor design for it not to be a constant, but we can't because the
> routines set/clear the LEDs use mode in-out.

Similarly,

    All_LEDs : GPIO_Points := Green & Orange & Red & Blue;

should be a constant.

pat-rogers:
> So I changed the modes in the original, since it allows use like the above and doesn't lose 
> anything since we aren't really changing the actual parameter anyway.

I don't mean that I have changed the current code, the HAL is still using in-out.

Fabien-Chouteau:
On Fri, May 13, 2016 at 8:02 PM, Pat Rogers <notifications@github.com>
wrote:

> On 5/13/2016 12:44 AM, Pat Rogers wrote:
> > On 5/12/2016 5:02 AM, Fabien wrote:
> >> I think I'm the one to blame. STM32.GPIO.GPIO_Point is now implementing
> the HAL.GPIO.GPIO_Point interface and I defined this interface with "in
> out" for the clear and set procedures.
> >>
> >> I kinda like the possibility to control whether or not a piece of code
> can change the state of a GPIO_Point.
> >
> > It isn't an easy decision, and I had it the same way as your interface
> > originally.
> >
> > The indirection via the access component makes it an odd case because we
> > aren't changing the point, we're changing the state of the GPIO port it
> > designates. So mode in-out looks like we are doing something we are not.
> >
> >> What is the reason for you to have the GPIO_Points constant?
> >
> > We want to say this, for example:
> >
> > Pattern : constant array (Index) of User_LED := (Orange, ..., Green);
> >
> > It is poor design for it not to be a constant, but we can't because the
> > routines set/clear the LEDs use mode in-out.
>
> Similarly,
>
> All_LEDs : GPIO_Points := Green & Orange & Red & Blue;
>
> should be a constant.
>

I agree, but also consider that this is a generic interface. It very
possible that an implementation of this interface will need to be mutable
because, for instance it stores the state of each IO.


-- 
Fabien Chouteau

pat-rogers:
> I agree, but also consider that this is a generic interface. It very
> possible that an implementation of this interface will need to be mutable
> because, for instance it stores the state of each IO.

Yup.

Out of curiosity, do you know of any like that?

Fabien-Chouteau:
On Fri, May 13, 2016 at 8:21 PM, Pat Rogers <notifications@github.com>
wrote:

> > I agree, but also consider that this is a generic interface. It very
> > possible that an implementation of this interface will need to be mutable
> > because, for instance it stores the state of each IO.
>
> Yup.
>
> Out of curiosity, do you know of any like that?
>

I'm working on an driver for external IO chip, storing the state of the IO
could avoid some SPI or I2C requests.

-- 
Fabien Chouteau

Fabien-Chouteau:
For this reason, I wonder if the best choice is for all subprograms of an
interface to have the first argument as "in out".
And have this as a rule.

-- 
Fabien Chouteau

pat-rogers:
>> Out of curiosity, do you know of any like that?
>
> I'm working on an driver for external IO chip, storing the state of the IO
> could avoid some SPI or I2C requests.

Interesting. Thanks.

pat-rogers:
On 5/13/2016 1:37 PM, Fabien wrote:
> For this reason, I wonder if the best choice is for all subprograms of an
> interface to have the first argument as "in out".
> And have this as a rule.

Probably too strict. I've never seen it be necessary before.

PR #15, opened 2016-04-13
Removed aspect "Pack", because it conflicts with aspect "Size"

elgor:
The aspect "Pack" will tell the compiler to compress the array, so it will choose the smallest possible representation. However we want a fixed size with the "Size" aspect, therefore aspect "Pack" conflicts with aspect "Size". Luckily, the compiler may (and actually does) ignore the aspect "Pack".

See: https://en.wikibooks.org/wiki/Ada_Programming/Pragmas/Pack#Exact_data_representation


Fabien-Chouteau:
Looks good.

Note that we are working on a new version (the svd branch) of the drivers using code generated from SVD hardware description format and with a better interface definition.

Thanks for your contribution!

PR #19, opened 2016-04-20
Component_Size instead of Pack

elgor:
Hi, sorry for the compiler error. I forgot to insert Component_Size. I still think that Pragma Pack is not the right way to do it and it is better to clearly specify the size although both methods produce the same result. 
I compiled the train example without errors.

I am currently working on my master thesis and I am planning to use this library. Is there any discussion forum or mailing list for the development? I would like to follow the development process.

### DIFF
--- a/ARM/STMicro/STM32/drivers/stm32f4.ads
+++ b/ARM/STMicro/STM32/drivers/stm32f4.ads
@@ -74,9 +74,9 @@ package STM32F4 is
    type Bits_29 is mod 2**29 with Size => 29;
    type Bits_30 is mod 2**30 with Size => 30;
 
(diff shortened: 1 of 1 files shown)


lambourg:
Hello Elgor,

We've just merged our long-term branch svd with the trunk and this data structure is not directly used anymore (so it is now obsolete). Basically, we now only rely on very basic integer definitions (from 1 bit integers, booleans, to 32 bit integers, words, and all intermediate values, as the form UIntX). The array definitions are now the responsibility mainly of the registers as defined by the SVD2Ada binding generator, and they should contain already the proper aspects.

So I'm not merging this pull request. Feel free to comment on the new architecture and continue to propose fixes when you see the need to.

Best regards,
- Jerome

PR #20, opened 2016-04-24
Svd

lambourg:
Time to move the SVD branch to the master branch.

In particular the overall architecture of the branch becomes more and more stable and makes now lots of sens. The different layers are now properly separated and identified:
- MCU
- Board
- HAL

Also, a last effort has been made to keep the compatibility with the GNAT GPL 2015, so the branch is useable "almost" as is (still requires a proper run-time for the different boards).


pat-rogers:
Done.

PR #21, opened 2016-04-28
Fixed violation of restriction "NO_FIXED_POINT"

elgor:
I tried to compile with `pragma Restrictions (No_Fixed_Point)` but got an error in 

drivers/i2c-stm32f4/stm32-i2c.adb : 392:13 
drivers/i2c-stm32f4/stm32-i2c.adb : 392:43

violation of restriction "NO_FIXED_POINT"

Instead of `To_Duration` I used the Time Type `Milliseconds` as it is used in other places within the file.

The comment at the implementation of `To_Duration` in a-retide.adb says:
```ada
   --  This function is not supposed to be used by the Ravenscar run time and
   --  it is not supposed to be with'ed by the user either (because it is an
   --  internal GNAT unit). It is kept here (returning a junk value) just for
   --  sharing the same package specification with the regular run time.

   function To_Duration (T : Time) return Duration is
      pragma Unreferenced (T);
   begin
      return 0.0;
   end To_Duration;
```

Furthermore the type `Natural` of `Timeout` is not specific enough, because the unit is not clear. Maybe it should be of type `Time` in Ada.Real_Time. Then one could call the function with `Timeout => Milliseconds(whatever)`


lambourg:
Hello Elgor,

This looks good, however there's a small style issue: subprogram calls should have a space between the subprogram identifier and the parameters:
Milliseconds(Timeout) should be Milliseconds (Timeout).

This style warning should have been spoted by GNAT normally. Once this is fix, I'll merge your proposal.

Fabien-Chouteau:
Hello @elgor ,

The function To_Duration that is used in stm32-i2c.adb comes from Ada.Real_Time (a-reatim) not 
Ada.Real_Time.Delays (a-retide), there's no problem to use Ada.Real_Time.To_Duration.

Ada.Real_Time.Delays (a-retide) is an internal GNAT unit, it means that user code should not with it. In fact if you try to with this unit there will be a warning `warning: "Ada.Real_Time.Delays" is an internal GNAT unit`.

Your patch is still correct and in fact it's in the procedure Wait_Flag above we use a call to Milliseconds instead for converting to duration. However can you re-start from a clean fork because there are 4 commits in your pull request that are not related to this fix.

Thanks,

PS: Why do you want to compile with `pragma Restrictions (No_Fixed_Point)`?

elgor:
I tried to get rid of the necessary commits with git rebase but I got an error. Is there a better way for this?

elgor:
Hi @Fabien-Chouteau,

I think I followed the goto declaration of Ada.Real_Time.To_Duration in GPS, so maybe that was misleading. I use parts of the library for my own project (master thesis) and this had the restriction pragma set.
However, I don't see the style checks because I use my own gpr files. I also tried to compile one of the examples in the complete library to check if the changes in the pull request will work for you but I could not compile it because of:

`missing argument for parameter "Device_ID" in call to "HSE_CLOCK" declared at s-bbpara.ads:75`

It seems that you use a different runtime than me. I ported my runtime for the smt32f427 from the GNAT GPL runtime according to the tutorial at http://blog.adacore.com/porting-the-ada-runtime-to-a-new-arm-board. 
I was able to compile the examples in the old master branch. Did you switch the runtime or is there something I have to adjust?

Thanks in advance

Fabien-Chouteau:
I also used GPS to find were To_Duration was implemented (CTRL+Left click on the function name), it's strange that you ended up in Ada.Real_Time.Delays...

You should always compile with the project files from the repository otherwise you will never see if your commits breaks the builds. There are few reasons to add new project files, for instance new device/board support or new example. If that is what you want to do please copy an existing project file and modify it to your need.

The run-time error you see is because HSE_CLOCK used to be a function and it's now a constant value. I'm sorry for the frustrating experience, we will be more careful in the future, starting with 2016 GPL release. We are also thinking of a solution to improve the "run-time" experience.

lambourg:
@elgor indeed, and that's my bad: I noted at some point in the blog article you're referring to that an "optional" optimisation to the runtime would be to make the HSE_Clock value a constant.

The rational being: the HSE clock value (High Speed External clock) being a board property, having a function returning some value depending on the MCU is wrong (it depends on the board, not on the MCU, and ST having decided to use 8MHz clock for all their STM32F4 disco boards is not enough to make it depends on the MCU, and not even true as the initial version of the STM32F469-disco board had a 25MHz crystal).

So now it's a constant value that is board specific (now meaning internally, and in the next GPL release), and does not depend on the MCU version.

In your custom runtime for the STM32F429-disco board, you just need to edit s-bbbopa.ads and replace HSE_Clock with a constant, fix the few places where in the runtime it is used as a function, and you should be OK to use the drivers.

elgor:
@lambourg, thank you for the advice, I was able to adjust the runtime according to your instructions. 
However two additional preparations were necessary (for people with the same problem):

 1. disable warnings for conditions (-gnatwC) in runtime_build.gpr:34

 `ALL_ADAFLAGS := ("-gnatpg", "-gnatn2", "-nostdinc", "-gnatwC") & CFLAGS;`

  otherwise the compiler will complain that some if conditions are always false because HSE_Clock is now constant.
 2. remove unsupported linker flag `--print-memory-usage` in 
		* examples/common/common.gpr:42
		* examples/accelerometer/accelerometer.gpr:14   (comment line)

	At least my GNAT GPL version didn't know this flag.

lambourg:
Thank you for the precisions. I will add those precisions in the Readme.md so that everyone can benefit from this, until the GNAT GPL 2016 is out.

PR #26, opened 2016-06-29
Fixed missing NAK for last 2 Bytes read

elgor:
I think I found a bug in I2C multibyte read.

If an even number of bytes > 2 is read, the receive method will end up in the `elsif Idx + 1 = Data'Last then` branch. However in there, the ACK is not disabled. After receiving all bytes the master will ACK the last byte and the slave will try to send more bytes but the master disables the Clock.

In my case the slave was pulling down SDA forever and the I2C bus was stuck.

### DIFF
--- a/ARM/STM32/drivers/i2c_stm32f4/stm32-i2c.adb
+++ b/ARM/STM32/drivers/i2c_stm32f4/stm32-i2c.adb
@@ -796,6 +796,8 @@ package body STM32.I2C is
 
          elsif Idx + 1 = Data'Last then
             --  Two bytes to read
+            Handle.Periph.CR1.ACK := False;
+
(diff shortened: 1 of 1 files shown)


Fabien-Chouteau:
Thanks for having a look at this. I noticed this problem the other day. The patch looks good.

BTW, do you know why we have to handle so many special cases:

         if Idx = Data'Last then
            --  One byte to read
            [...]
         elsif Idx + 1 = Data'Last then
            --  Two bytes to read
            [...]
         elsif Idx + 2 = Data'Last then
            --  Three bytes to read
            [...]
         else
            [...]
         end if;

I seems like we could only have two cases:

         if Idx = Data'Last then
            --  One byte to read
         else
            [...]
         end if;

PR #33, opened 2016-08-09
added DMA write to SD card drivers

mbeckersys:
Successfully tested on a STM32F427VIT and merged back into this library for public use.

The TX_Active bit is hanging around for a bit, leaving a single DMA transfer to be a bit slow, yet (~2msec).

PR #34, opened 2016-08-10
bugfix for broken STM32F7 build after pull request #33

mbeckersys:
Fixes the broken STM32F7 build.

### DIFF
--- a/ARM/STM32/drivers/sd/sdmmc/stm32-sdmmc.ads
+++ b/ARM/STM32/drivers/sd/sdmmc/stm32-sdmmc.ads
@@ -24,6 +24,7 @@ package STM32.SDMMC is
       Unsupported_Card,
       Rx_Overrun,
       Tx_Underrun,
+      Startbit_Not_Detected,
       Request_Not_Applicable,
       CRC_Check_Fail,
       Illegal_Cmd,
@@ -44,7 +45,8 @@ package STM32.SDMMC is
       WP_Erase_Skip,
       Erase_Reset,
       AKE_SEQ_Error,
-      Invalid_Voltage_Range);
+      Invalid_Voltage_Range,
+      DMA_Alignment_Error);
 
    type Supported_SD_Memory_Cards is
      (STD_Capacity_SD_Card_V1_1,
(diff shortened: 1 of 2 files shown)


lambourg:
Hello Martin,

I already committed a fix with more or less the same changes. Can you try it out to see if this still works for you?

The thing is I integrated recently changes in the SVD description file for the stm32f7x line of MCUs to match what's done on the stm32f769 one which is more recent and does things a bit differently on the sdio/sdmmc front... This change renames the interrupt to SDIO_Interrupt (same as the stm32f4 line), but this is not visible yet as this is part of the runtime. This also changes the SDMMC1 peripheral type as SDMMC, that you caught already in your patch.

mbeckersys:
Hi Jérôme,

I only have STM32F427 devices here and my pull request was based on a clone just a few hours old at that point. Thus, the only changes since then which I _could_ test would be those of 35cd551 in  media_reader-sdcard.adb, which are the same for the STM32F7 line. Unless I have missed something.

I can do that nevertheless, but it will take me a few days.

lambourg:
No problem, there's no hurry. At least the example compiles for now and it's still very early stage support for sdcards (I know of many problems with it that are still around).

mbeckersys:
It might be early stage, but for us it is working on the STM32F427. We 
are writing to a FAT file system without any problems. Just the 
performance is a bit low, limited by the lingering TX_ACTIVE bit of the 
SDIO peripheral.

By the way, I have accidentally left a few comments in the write 
procedure, that suggest otherwise. Everything is working...so maybe 
delete them. There are from before the DMA FIFO was setup properly.

I was also thinking about contributing our FAT write implementation, but 
you are so fast, that in those few weeks our FAT codes diverged quiet a bit.

Am 11.08.2016 um 23:49 schrieb Lambourg Jérôme:

lambourg:
Ok, so that's good news :)

Well, if you have working code in this area, please contribute. It's been quite some time I wanted to continue this implementation, but couldn't find time. Maybe a branch would be OK so that I try to merge things here.

Note that this Ada FATFS library that I started in the example is intended to be LGPL (e.g. GPL library license: you can't distribute a modified binary version of it unless it remains in the same license terms, but you can use the library as-is in closed-source products). This is opposed to the rest of the drivers library that uses a lesser copyleft license.

PR #36, opened 2016-08-15
Adafatfs write capabilities

mbeckersys:
This is a pull request for my adafatfs write implementation, tested successfully on STM32F427. Capabilities: create directories, create files, write to files, overwrite files and some related things. Names so far limited to 8+3 (i.e., no VFAT entries are created).

This request is split into two commits:
 1. 7100ec8: a backport of the recently merged pull request #33 (SD DMA write) to the branch point
 2. 7321bbd: patches to adafatfs to add write capabiities

The first part can be ignored when merging, since this is all contained in master's HEAD already (these are the relevant patches that I needed to set up this pull request). The second part needs manual merging, since our two adafatfs implementations diverged to a fair amount.

It certainly could use some refactoring and optimization, but it works reliably so far.

### DIFF
--- a/ARM/STM32/drivers/sd/sdio/stm32-sdmmc.ads
+++ b/ARM/STM32/drivers/sd/sdio/stm32-sdmmc.ads
@@ -24,6 +24,7 @@ package STM32.SDMMC is
       Unsupported_Card,
       Rx_Overrun,
       Tx_Underrun,
+      Startbit_Not_Detected,
       Request_Not_Applicable,
       CRC_Check_Fail,
       Illegal_Cmd,
@@ -44,7 +45,8 @@ package STM32.SDMMC is
       WP_Erase_Skip,
       Erase_Reset,
       AKE_SEQ_Error,
-      Invalid_Voltage_Range);
+      Invalid_Voltage_Range,
+      DMA_Alignment_Error);
 
    type Supported_SD_Memory_Cards is
      (STD_Capacity_SD_Card_V1_1,
@@ -64,7 +66,7 @@ package STM32.SDMMC is
       Data_Read_Access_Time_2          : Byte; --  In CLK Cycles
       Max_Bus_Clock_Frequency          : Byte;
       Card_Command_Class               : Short;
-      Max_Read_Data_Block_Length       : Byte;
+      Max_Read_Data_Block_Length       : Byte; -- ld (blocksize in bytes)
       Partial_Block_For_Read_Allowed   : Boolean;
       Write_Block_Missalignment        : Boolean;
       Read_Block_Missalignment         : Boolean;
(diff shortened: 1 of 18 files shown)


lambourg:
Thank you very much for this pull request. I'll try to merge all that :) I'll keep you in touch as soon as this is the case.

Fabien-Chouteau:
Thanks Martin! 

I've started a "file_system_and_block_drivers" branch were I plan to experiment on a few things. I think we need some HAL interfaces to handle to abstract the SD card and FATfs implementation:
 - Block driver: Reading and writing blocks of 512 bytes. Implementation could be SD card, RAM, SPI/I2C EEPROM, USB mass storage.
 - SD card driver: extending the block driver with SD specific info (card id, card info, etc.)
 - File system: inspired by the POSIX and libfuse interfaces. Implementations could be FATfs, virtual fs to handle mounting different FS, RAM FS, crypto FS.

Beyond allowing support for file system on more MCUs, having those interfaces will also mean that we will be able to test the file system implementation on host machines (Windows, Linux) with a file base block driver mock-up.

If you guys want to join in :)

mbeckersys:
Dear Fabien,

That sounds like a good idea. What do you think of wiring this stack in SPARK, as far as possible? I found it tricky to get all the bits right, especially when thoughts of optimization come into play. 
We could then see how far we can go with optimizing, without loosing grip in the analysis. Any maybe push the envelope of the analysis a bit.

lambourg:
On the pure driver front, I already started a mockup of abstraction in the sdcard example with the 'media_reader' layer: the goal was also to be able to plug the board's EEPROM there.

Fabien-Chouteau:
> That sounds like a good idea. What do you think of wiring this stack in SPARK, as far as possible? 

It's not a goal of this project per se, but if/when it makes sense I'm OK with SPARK.

> I found it tricky to get all the bits right, especially when thoughts of optimization come into play.

Do you have examples of why optimization are tricky with SPARK?

Fabien-Chouteau:
> On the pure driver front, I already started a mockup of abstraction in the sdcard example with the 'media_reader' layer: the goal was also to be able to plug the board's EEPROM there.

As discussed live, I missed that. It's what I have in mind with the block driver (Linux terminology).

lambourg:
Hello Martin,

This is to let you know that I worked a bit on the FAT FS support in the sdcard example, and most functionalities are now there, even though this would now require some refactoring and polishing, as well as a merge with Fabien's work on the HAL filesystem API.

Basically, you can now create/read/delete directories, and add or delete entries to those (files or subdirs).
You can now also open files in r/w/rw mode.

There are still some missing functions, such as file seek, renaming, and also I tried to test things a bit, it is certainly still far from stable.

Anyway, you should be able to start playing with it, and send me feedbacks if you have trouble with the implementation.

Regards,
- Jerome

PR #38, opened 2016-08-17
Re-introduce the Cortex-M layer of projects.

lambourg:
This makes explicit the MCU family


Fabien-Chouteau:
We have to find a better solution for all those obj and lib directories. It's getting ridiculous :)

lambourg:
> Le 17 août 2016 à 18:44, Fabien Chouteau <notifications@github.com> a écrit :
> 
> We have to find a better solution for all those obj and lib directories. It's getting ridiculous :)

Simple one would be for gprbuild to create directories by default :D

Fabien-Chouteau:
> Simple one would be for gprbuild to create directories by default :D

Let's try to make it happen :)

pat-rogers:
On 8/17/2016 11:44 AM, Fabien Chouteau wrote:
> We have to find a better solution for all those obj and lib directories. It's getting ridiculous :)


Do we also still have the problem in which a fair amount of code is 
recompiled every time a build is invoked?

lambourg:
> Le 17 août 2016 à 20:46, Pat Rogers <notifications@github.com> a écrit :
> 
> On 8/17/2016 11:44 AM, Fabien Chouteau wrote:
> > We have to find a better solution for all those obj and lib directories. It's getting ridiculous :)
> 
> 
> Do we also still have the problem in which a fair amount of code is 
> recompiled every time a build is invoked?

Normally, with my change in project organization, it's all fixed in the library itself.

Note that those were recompiled only upon runtime or build type change (or
different board used).

Now within the library, all those configuration have dedicated folders, hence
the huge number of those.

PR #39, opened 2016-08-17
Move alternate functions from STM32.GPIO to STM32.Device.

lambourg:
Alternate functions are device-specific


Fabien-Chouteau:
I'm not a fan of adding the value of AF within it's name.

lambourg:
With recent MCUs (STM32F469, F769 in particular), very often devices are spread across several AF functions, leaving ambiguity as to how to configure a given pin.

And to properly configure the pins, you have to have a look at the datasheet that references the AF number, so have a reminder here is good IMHO.

Fabien-Chouteau:
A couple things I missed:

 - examples/sdcard/sdcard_stm32f769disco.gpr should not be part of this commit:
 - Enable_Clock and Rest should be in stm32-device not examples/sdcard/src/stm32f769/device_sd_configuration.adb (and also not part of this commit)
 - You not only move alternate functions from STM32.GPIO to STM32.Device, you also introduce a new type GPIO_AF_Point. I don't think we need yet another GPIO type.

lambourg:
stm32f769 could indeed have been commited separately. Concerning Enable_Clock and Reset, it's there in the examples as long as the support for the sdcard is not ready for prime time. There's still many bugs to fix there.

The GPIO_AF_Point is to address the new devices I talked about. The LTDC on the STM32F769 spans 3 or 4 AF functions, so having a convenient way to configure the pins is IMHO interesting. We could for example using such configuration record properly document the pins of a given MCU, and the various options to configure it.

Fabien-Chouteau:
> stm32f769 could indeed have been commited separately.

Yes, let's have clean commit.

> Concerning Enable_Clock and Reset, it's there in the examples as long as the support for the sdcard is not ready for prime time. There's still many bugs to fix there.

We have a lot of things that don't work, I would recommend putting it at the right place first. And those are simple sub-programs that will not change and that we know where to put.

> The GPIO_AF_Point is to address the new devices I talked about. The LTDC on the STM32F769 spans 3 or 4 AF functions, so having a convenient way to configure the pins is IMHO interesting. We could for example using such configuration record properly document the pins of a given MCU, and the various options to configure it.

It's indeed a interesting subject consider but it doesn't belong in this commit. We need to discuss that and how to handle the giant number of combinations. (this cannot be generated from SVD right?)

lambourg:
> stm32f769 could indeed have been commited separately.
> 
> Yes, let's have clean commit.
> 
> Concerning Enable_Clock and Reset, it's there in the examples as long as the support for the sdcard is not ready for prime time. There's still many bugs to fix there.
> 
> We have a lot of things that don't work, I would recommend putting it at the right place first. And those are simple sub-programs that will not change and that we know where to put.
> 
> The GPIO_AF_Point is to address the new devices I talked about. The LTDC on the STM32F769 spans 3 or 4 AF functions, so having a convenient way to configure the pins is IMHO interesting. We could for example using such configuration record properly document the pins of a given MCU, and the various options to configure it.
> 
> It's indeed a interesting subject consider but it doesn't belong in this commit. We need to discuss that and how to handle the giant number of combinations. (this cannot be generated from SVD right?)

Right, there's no such information in the SVD file. I'll have a look if some parseable file can be found at ST, but I doubt that.

For reference, this interesting work on the same concern (see first answer): http://nuttx.yahoogroups.narkive.com/bl17mJws/help-wanted-review-stm32l-pinmaps

Fabien-Chouteau:
As I expected, it's a big giant enormous amount of work for a low added value IMHO.

Unless ST opens the CubeMX tool, I don't think it's worth putting more effort into this.

pat-rogers:
I agree that the constants should have the underlying numeric values as part of the identifiers since some names are otherwise identical. I also agree that we should have just this one change in this commit.
That said, it would be good to retain the protection against invalid values that the private type currently provides. Otherwise users are likely to just start using numeric literals at some point, rather than our constants, which is clearly more prone to errors. So to handle the fact that the values are device-specific, what I had thought about doing was to have a single common private type in a parent package, with deferred constants in device-specific child packages. Thoughts?

Fabien-Chouteau:
> That said, it would be good to retain the protection against invalid values that the private type currently provides. Otherwise users are likely to just start using numeric literals at some point, rather than our constants, which is clearly more prone to errors. So to handle the fact that the values are device-specific, what I had thought about doing was to have a single common private type in a parent package, with deferred constants in device-specific child packages. Thoughts?

Good point, I like that.

lambourg:
I finally merged manually with all your comments: minimal change, just move the AF definitions from the stm32.gpio package to the stm32.device package, keeping the renaming with the AF number.

I'm closing this merge request then.

pat-rogers:
Sweet.

PR #45, opened 2016-08-30
File system and block drivers

Fabien-Chouteau:
@mbeckersys , @lambourg this is the result of my file system interface experimentation.

Here's a list of the new features:
- Services library: Stuff that implement HAL interfaces but are not physical components. In the future: networking, USB, Bluetooth, simple Bitmap drawing, logging system, etc.
- HAL.Filesystem: File system interface.
- HAL.Block_Drivers: Interface to read/write in disk, mass storage, etc.
- Virtual_File_System: Mount a file system within a file system
- Semihosting.Filesystem: reading files on the host as if they where on the target FS (through ARM semihosting)
- Partitions tool: To read partitions entries on a disk
- File_Block_Driver: Simulate a disk by reading into a file. When mixed with Semihosting.Filesystem, this means we can test FS implementations by manipulating host files.


hal/src/hal-filesystem.ads
```diff
@@ -0,0 +1,145 @@
...
+                                Permissions : Permission_Set)
+                                return Status_Kind is abstract;
+
+   function Change_Owner_And_Group (This  : in out FS_Driver;
+                                    Path  : Pathname;
+                                    Owner : User_ID;
+                                    Group : Group_ID)
+                                    return Status_Kind is abstract;
+
+   function Change_Permissions (This   : in out FS_Driver;
```
lambourg: I don't see the relationship between the name of this subprogram and its parameters...

hal/src/hal-filesystem.ads
```diff
@@ -0,0 +1,145 @@
...
+                    Old_Path : Pathname;
+                    New_Path : Pathname)
+                    return Status_Kind is abstract;
+
+   function Change_Permissions (This        : in out FS_Driver;
+                                Path        : Pathname;
+                                Permissions : Permission_Set)
+                                return Status_Kind is abstract;
+
+   function Change_Owner_And_Group (This  : in out FS_Driver;
```
lambourg: In the context of a bare metal FS driver, I think this one is overkill...

hal/src/hal-filesystem.ads
```diff
@@ -0,0 +1,145 @@
...
+
+package HAL.Filesystem is
+
+   subtype Pathname is String;
+
+   type File_Kind is (Regular_File, Directory);
+
+   type File_Mode is (Read_Only, Write_Only, Read_Write);
+
+   type Permission_Kind  is (Others_Execute, Others_Write, Others_Read,
```
lambourg: This is highly FS-specific. In general, and in the context of Bare Metal FS, I'm not sure we'll need that. Then if we really want full support for permissions, most modern FS also support advanced ACL that go far beyond this definition.

My take on this is that all we need is queries on whether the application has read or write access to a specific path (I don't see us using the 'execution' access).

Fabien-Chouteau: I don't agree that access control is overkill for the Ada_Drivers_Library, it's key feature of file systems, even if older ones don't implement it. But I don't mind leaving that out for the moment.

hal/src/hal-filesystem.ads
Fabien-Chouteau: That's because it's a typo...

Fabien-Chouteau:
I fixed the typo and removed ownership and access rights primitives,

PR #46, opened 2016-08-31
fix build problem due to case issue

VictorLSogilis:
Hi,

We tried to build the latest version of certyflie but we encounter build issue due to case problems. I suppose it works on a case insensitive file system (FAT Windows ?) but on Linux it's quite uncommon.

### DIFF
--- a/boards/crazyflie.gpr
+++ b/boards/crazyflie.gpr
@@ -18,7 +18,7 @@ aggregate library project Crazyflie is
 
    for Project_Path use ("OpenMV2");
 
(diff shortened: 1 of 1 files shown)


Fabien-Chouteau:
Thanks!

Fabien-Chouteau:
Actually the line above is wrong as well:
` for Project_Path use ("OpenMV2");`
should be  
`for Project_Path use ("crazyflie");`

@VictorLSogilis, do you mind fixing this?

PR #47, opened 2016-08-31
fix project path for crazyflie

VictorLSogilis:
side fix of project path as suggested in https://github.com/AdaCore/Ada_Drivers_Library/pull/46


Fabien-Chouteau:
Thanks Victor! :)

PR #52, opened 2016-09-08
Correction

astro-jerry:
Line 422 : I2C3_Base should be changed to I2C4_Base


Fabien-Chouteau:
Thanks Jerry! Did you check if this error is also present other stm32-device.ads?

astro-jerry:
Yes, I checked the others but it was only in the F7 file.

Fabien-Chouteau:
Great! Thanks again

PR #64, opened 2016-10-01
Rename Word to UInt32 and Short to UInt16

Fabien-Chouteau:
The use of Word and Short to designate data type is ambiguous, not only
they do not express the signed or unsigned nature of the type, the size
is also different depending on the architecture. We might support 64bit
machines in the near future so it's time to change. Using UInt32 and
UInt16 is also more coherent with the other unsigned integer types used
all across the library.

This patch also define UIntXX up to 64.


pat-rogers:
Agreed!

PR #67, opened 2016-10-10
First stage of Cortex-M0, nRF51 and MicroBit support

Fabien-Chouteau:
This patch series provides a first batch of drivers for the Nordic nRF51. It includes:
 - Clock
 - Events
 - GPIO
 - GPIO Tasks and Events
 - Interrupts
 - PPI
 - RNG
 - RTC
 - Tasks
 - Temperature
 - TWI (I2C)

It also adds support for the ARM Cortex-M0 NVIC as well as the MicroBit board that uses the nRF51.


pat-rogers:
Looks good.

PR #69, opened 2016-10-16
Start native implementations for HAL interfaces and start a native testsuite

pmderodat:
Hi,

This patchset fixes the “native” board and adds to it an implementation for HAL.Filesystem, based on standard Ada packages so that we can run services/components that use this interface on a native target. This makes it possible to run a testsuite (with a native setup).

Currently, the testsuite runner is a custom but simple Python script but I guess we could move to something else easily if we find better. :-)


Fabien-Chouteau:
Thanks @pmderodat ! Tackling quality assurance like a boss :)

At first Python seems like a burden for Windows users, but there's a Python in every GPS packages so it should be OK. 

I will have a closer look at the patches.

pmderodat:
Haha, thanks. ;-)
I had no idea Python was an issue on Windows. GPS embeds a Python2 interpreter while my script uses Python3, so if needed I can tweak it to run on Python2 instead. Otherwise, I’m not sure if an Ada runner would be less a burden: I’ve always had problem to properly deal with subprocesses in Ada.

Anyway I’m totally open to other ideas.

hal/src/hal-filesystem.ads
```diff
@@ -116,12 +153,22 @@ package HAL.Filesystem is
...
 
    function Entry_Name (This         : in out Directory_Handle;
                         Entry_Number : Positive)
                         return Pathname is abstract;
+   --  Return the absolute path (full name) for the Entry_Number directory
+   --  entry. Return an empty path if this entry does not exist.
+   --
+   --  ??? In a virtual file system, this full path will actually be relative
+   --  to the root directory of the relevant file system. This makes VFS
+   --  traversal cumbersome to write.
```
Fabien-Chouteau: I don't think we want to return the full path here, but just the name of the node.

Fabien-Chouteau:
> I had no idea Python was an issue on Windows.

Well it's not as easy as Linux where it's always installed in your distribution. Except for GPS, I have no Python install on my Windows machine.

> GPS embeds a Python2 interpreter while my script uses Python3, so if needed I can tweak it to run on Python2 instead.

That would be great.

> Otherwise, I’m not sure if an Ada runner would be less a burden: I’ve always had problem to properly deal with subprocesses in Ada.

I really don't know, to me the key point is ease of use.

hal/src/hal-filesystem.ads
```diff
@@ -92,17 +120,26 @@ package HAL.Filesystem is
...
+   --  Write bytes in Data to This. This overrides the next Data'Length bytes in
+   --  this file, if they exist, it extends the file otherwise.
 
    function Seek (This   : in out File_Handle;
                   Offset : IO_Count)
                   return Status_Kind is abstract;
+   --  Set the read/write cursor of This to point to its byte at the absolute
+   --  Offset.
+   --
+   --  ??? What should happen if this offset is beyond the end of the file?
```
pmderodat: Yeah but does this actually extend the file? Or does that happen only at the next write?

hal/src/hal-filesystem.ads
pmderodat: Seems more sensible to me thanks. :-) I’ll look for the place that made me think it returned the full path.

services/src/filesystems/virtual_file_system.adb
```diff
@@ -61,9 +63,40 @@ package body Virtual_File_System is
                     Path       : Pathname)
                     return Status_Kind
    is
-      pragma Unreferenced (This, Path);
+      procedure Destroy is new Ada.Unchecked_Deallocation
```
pmderodat: Arf, I just noticed that it’s actually a bad idea to do that: it breaks the build of some examples with link errors such as:
`[…]/Ada_Drivers_Library/services/src/filesystems/virtual_file_system.adb:90: undefined reference to `__gnat_free'`

Ok, so the runtime we use for stm32f407disco provides `__gnat_malloc` but it does not provide `__gnat_free`. This leaves me wondering: is it legitimate in this setup to dynamically *allocate* memory?

hal/src/hal-filesystem.ads
Fabien-Chouteau: ah, I didn't know POSIX lseek could change the size of the file... I guess something closer to UNIX lseek would be less confusing, with SEEK_SET, SEEK_CUR, SEEK_END.

services/src/filesystems/virtual_file_system.adb
Fabien-Chouteau: Ravenscar-full has __gnat_free, but no Ravenscar-sfp. As I said we should provide both heap and pool allocation policies.

testsuite/tests/virtual_file_system/tc_empty_vfs.out
```diff
@@ -0,0 +1,4 @@
+Dumping the root directory of an empty VFS:
```
Fabien-Chouteau: We can't have empty directories in a git repository. So maybe the test should create that directory before.

pmderodat: Ah, you’re right! I’ll do that, thanks!

PR #71, opened 2016-10-21
STM32: Rename Enable_Clock procedures to Power_Up

Fabien-Chouteau:
Even if the name Enable_Clock is accurate with regards to the
documentation and the internal behavior of the MCU, it is not clear at
all from a user point of view.

This patches use the name Power_Up which describes better the actual use
a these procedures.

Note: This is a partial commit to propose the idea. If it is well
received I will rename the remaining Enable_Clock.

### DIFF
--- a/ARM/STM32/devices/stm32f42x/stm32-device.adb
+++ b/ARM/STM32/devices/stm32f42x/stm32-device.adb
@@ -56,7 +56,7 @@ package body STM32.Device is
    -- Enable_Clock --
    ------------------
 
-   procedure Enable_Clock (This : aliased in out GPIO_Port) is
+   procedure Power_Up (This : aliased in out GPIO_Port) is
    begin
       if This'Address = GPIOA_Base then
          RCC_Periph.AHB1ENR.GPIOAEN := True;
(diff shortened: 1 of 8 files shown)


pat-rogers:
I disagree, because the name indicates directly what it does, and because it corresponds to the examples in C provided by the vendor, so changing it would be misleading, and we want the existing knowledge expressed in other languages to be as applicable to Ada code as possible.

astro-jerry:
I have to agree with Pat on this.  I think that Enable_Clock expressed exactly what the hardware does and matches C usage that has been in place for a long time.  I don't think we want to make Ada even more different from C when it is not necessary. We want to win over C users not drive them away.

Fabien-Chouteau:
Fair enough, I don't have a strong opinion on this one, I was proposing this change because someone didn't not understand what Enable_Clock() was doing (I don't remember if it was here or on IRC). 

@astro-jerry, we are already very different from the C drivers. I don't think it's a problem because it is somewhat one of the objectives of this project, show how Ada can be a better language for device drivers. Also I think that users will always look at the example to see how things are done, so as long as examples uses Power_On() they will understand.

PR #72, opened 2016-11-02
Move the train example to a separate repository

Fabien-Chouteau:
This example doesn't bring any useful info compared to "draw" and
therefore adds an unnecessary maintenance effort to the
Ada_Drivers_Library project.

Others examples like Wolf, 2048, fractals, conway or balls should also
be moved.

A link to the new repository is added to the README.


pat-rogers:
I agree with the intent, but am concerned about fragmentation -- ie, many little examples, lost among the bigger repos. Maybe a single repo for all those proposed to be moved?

Fabien-Chouteau:
> I agree with the intent, but am concerned about fragmentation -- ie, many little examples, lost among the bigger repos. 

There's the list of projects in the README that will help with that.

> Maybe a single repo for all those proposed to be moved?

Having them all together means more maintenance work because they will have to use the same version of Ada_Drivers_Library, which means updating for one implies updating for all.

pat-rogers:
We should consider the train example separately from the others you mention.  The train example is much more valuable than the others you mentioned because it is useful for explaining how SPARK can prove abstract properties at conferences, and for the conference booth itself. What I am concerned about is that it will suffer from bit-rot to the point that it will be a lot of work to build it when it is needed, whereas if we maintain it incrementally there will be relatively little, easy work at any given library update.  So "thinking out loud" my suggestion is to keep it in the drivers library repo and move those others.

pat-rogers:
I'm still unconvinced that it should go, for the reasons mentioned earlier.

Fabien-Chouteau:
As discussed last we can accept small demo under the AdaCore account. I will create the repo and move it there.

pat-rogers:
OK, sounds good. Thanks.

Fabien-Chouteau:
@pat-rogers @lambourg I moved the demo to an AdaCore repo: https://github.com/AdaCore/SPARK_Railway_Simulation_Demo

PR #73, opened 2016-11-03
Remove the demos that are not actual examples.

lambourg:
As discussed...


pat-rogers:
Let's discuss this live.

pat-rogers:
It is not clear to me that all of these should go.  Some, yes, and 
perhaps others, but not all yet.

lambourg:
Current list of demos moving out:
- balls (does not bring much over draw, in particular does not use the touch screen)
- 2048: too complex, not an example
- conway: also too complex
- fractals: ditto
- sdcard: ditto, wav reader is not really an example of the drivers usage.
- wolf: ditto

We'll be missing an example of audio driver usage though (as has always been the case until I implemented this wav reader). Having a simple sinusoid generator and playing the sound is enough I guess.

pat-rogers:
I'm OK with removing them as long as they are put somewhere such that they are not lost altogether. Another project within github.com/AdaCore?

lambourg:
Sorry, I messed with my branches, currently under fix

pat-rogers:
On 12/16/2016 10:01 AM, Fabien Chouteau wrote:
> Fabien-Chouteau requested changes on this pull request.
>
> There too many commits that have nothing to do with the subject of this PR.


Are they not all deletions?

pat-rogers:
On 12/16/2016 10:05 AM, Lambourg Jérôme wrote:
> Sorry, I messed with my branches, currently under fix

Oh, OK thanks.

PR #75, opened 2016-11-06
BNO055 9DOF IMU driver

pat-rogers:
Insert abstract base type for BNO055 IMU and an I2C-based instance


components/src/motion/bno055/adafruit_bno055.ads
```diff
@@ -0,0 +1,639 @@
...
+
+--  The AdaFruit BNO055 breakout board supports either I2C or USART
+--  communication. Therefore, the primary type is abstract because the I/O
+--  facility will be implemented by concrete subclasses.
+
+pragma Restrictions (No_Streams);
+
+with System;
+with Interfaces;    use Interfaces;
+with Ada.Real_Time; use Ada.Real_Time;
```
Fabien-Chouteau: You cannot have a dependency on Ada.Real_Time because this makes the entire components layer unusable for boards that only have ZFP run-time available such as the MicroBit.

I know it's not perfect, but please use HAL.Time.

pat-rogers:
On 11/6/2016 4:10 PM, Fabien Chouteau wrote:
> Fabien-Chouteau requested changes on this pull request.
>
> I'm not sure about the AdaFruit prefix here, this driver can work for any BNO055 chips, not only the ones provided by AdaFruit.

Agreed. My demo uses something specific to the board but the driver need 
not do so.  I've made those changes.

> +package AdaFruit_BNO055_I2C is
> +
> +   type BNO055_IMU (Port : I2C_Port_Ref; Device : I2C_Address) is
>
> As far as I can the BNO055 can only have 0x29 or 0x28 as I2C address. Device should be of type UInt1, a boolean or an enum with Primary_Address and Alternative_Address values.

Yes, only two. I take your point but prefer to provide them (as 
constants) and let the user choose. That way future addresses are 
allowed.  Your change is the kind of thing I like to do -- prevent the 
possibility of mistakes -- but the risk doesn't seem worth the slightly 
heavy approach in this case.


> +with Ada.Real_Time; use Ada.Real_Time;
>
> You cannot have a dependency on Ada.Real_Time because this makes the entire components layer unusable for boards that only have ZFP run-time available such as the MicroBit.
>
> I know it's not perfect, but please use HAL.Time.

Understood.  Early on, I had concluded that not every driver would be 
supported by every runtime, and so I want to think about this some more. 
  One option would be to support a very much cut-down package for 
Ada.Real_Time, supporting (say) only relative delays.

PR #80, opened 2016-11-12
Make PWM abstract type simpler and more robust. Add additional new demo for PWM ADT.

pat-rogers:
Timers can still be shared among modulator objects.
Add new demo for using the abstraction (in addition to existing demo that used the timer directly).


Fabien-Chouteau:
Better interface indeed, the only drawback is that the frequency changed from Float to UInt32 but I'm not sure if sub 1Hz frequency is doable/desirable...

I also added a fix to the OpenMV support that was broken with your patch.

pat-rogers:
> I also added a fix to the OpenMV support that was broken with your patch.

Thanks for that.

PR #86, opened 2016-11-18
Testsuite improvements (verbose switch and no test error)

Fabien-Chouteau:


testsuite/run.py
```diff
@@ -139,12 +147,22 @@ def find_testcases():
 )
 
 parser.add_argument(
+    '--verbose', action='store_true',
+    help='Use Valgrind to detect invalid memory operations and leaks'
```
pmderodat: I have the feeling that you forgot to update the help message. ;-)

testsuite/run.py
```diff
@@ -139,12 +147,22 @@ def find_testcases():
...
+parser.add_argument(
     'pattern', nargs='*',
     help='List of pattern to filter the set of testcases to run'
 )
 
 
 def main(args):
+
+    if args.verbose:
+        global verbose
```
pmderodat: So far, I tried to avoid globals: I just passed “args” to all functions that need to inspect command-line options. Would it be possible to add it as well to the “run_program” function, instead?

testsuite/run.py
Fabien-Chouteau: Indeed... :grimacing:

testsuite/run.py
Fabien-Chouteau: Actually it's already done. I didn't have to create this global...

Fabien-Chouteau:
New version is up!

PR #87, opened 2016-11-19
MCP23x008: Move the IO primitives to the private part of the specs

Fabien-Chouteau:
And implement MCP23003 (I2C version) in a child package: MCP23x008.I2C.

### DIFF
--- a/components/src/io_expander/MCP23xxx/mcp23x08-i2c.adb
+++ b/components/src/io_expander/MCP23xxx/mcp23x08-i2c.adb
@@ -31,7 +31,7 @@
 
 with HAL.I2C; use HAL.I2C;
 
-package body MCP23008 is
+package body MCP23x08.I2C is
 
    --------------
    -- IO_Write --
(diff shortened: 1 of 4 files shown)


pat-rogers:
There are two different chips (MCP23008 and MCP23S08) involved, with slightly different capabilities, right?
If so, then I think type extension is indeed a good approach. OTOH, if the only difference is the internal communication mechanism, then I think a generic is better.

Fabien-Chouteau:
They have the same capabilities, just a different interface. It's probably the same silicon inside.

pat-rogers:
On 11/22/2016 3:41 AM, Fabien Chouteau wrote:
> They have the same capabilities, just a different interface. It's probably the same silicon inside.

But they have different part numbers, I believe, so the vendor thinks of 
them as distinct, and therefore separate types for them (via type 
extension) make sense.  I think the BNO055 is different in that regard.

Fabien-Chouteau:
Sounds good!

PR #89, opened 2016-11-21
Start support of nRF51 radio and Bluetooth Low Energy

Fabien-Chouteau:


Fabien-Chouteau:
> I've commented on functions named Get_* elsewhere

I can fix that :)

Fabien-Chouteau:
Done

PR #96, opened 2016-12-16
Revert "Remove the demos that are not actual examples."

Fabien-Chouteau:
Reverts AdaCore/Ada_Drivers_Library#95

The link in the README is wrong and the FATFS support has to be moved to the service layer instead of being removed.


lambourg:
The current version that is implemented in the example just does not work. I suggest that if you want to add this support to the service layer, you should start from the version that is in my repository then, as at least I could check that the write operations are working (and fixed many other glitches), and on which I performed some heavy refactoring (to simplify its architecture, put more code in common). Though this is still far from complete, in particular the allocation of handlers is still not satisfactory IMHO.

lambourg:
BTW I don't know why you closed the pull request to the fix to the README ? You're telling me that it's wrong!

PR #98, opened 2016-12-17
Correct DMA config usage error. More update for structural change for gpr file

pat-rogers:
Correct peripheral burst size (in config record) so that no data are lost.
Otherwise, three chars are lost on each transfer: instead of "abcdefghijk..." we
would get "dhlpt..."

Rename dirs for more accurate indication of demo purpose (DMA, not USART demos).

stm32-dma.ads: minor (valuable) formatting for readability.


pat-rogers:
Thanks

PR #99, opened 2016-12-17
Change name of user button package 

pat-rogers:
To reflect that it is not a general user-defined button. Misc readability name changes and formatting.
Update examples applying the user button for package name change.

Change name of GPIO function returning the external interrupt line number.


pat-rogers:
Thanks

PR #100, opened 2016-12-17
Ensure all demos have readme files. Remove redundant demo_dma_interrupts project.

pat-rogers:
Minor formatting within some of the code.


lambourg:
Now that the large examples have been moved out of the "examples" directory, I'd move those unitary examples into the root-level example directory. I don't really see any reason to make a difference between those and the hello_blinky one... Apart from that, looks good to me :)

pat-rogers:
What has kept me from suggesting moving them up is that they are STM-specific, but also that they are meant to be very narrow in scope.  Not sure what's best there.

lambourg:
have an stm32 subdirectory there?

PR #101, opened 2016-12-17
Unify access to classwide type names

pat-rogers:
An access-to-classwide type is much like a pointer-to-baseclass in C++, in that the access values can designate any type in the derivation class. **Any** type. A good way to name such types then, and one that does not redundantly include the "_Access" or "_Ref", is to add the prefix "Any_" to the designated type name, eg Any_GPIO_Point or Any_I2C_Port. If we are consistent the reader will immediately know what they are getting.

### DIFF
--- a/ARM/cortex_m/src/semihosting-filesystem.adb
+++ b/ARM/cortex_m/src/semihosting-filesystem.adb
@@ -126,7 +126,7 @@ package body Semihosting.Filesystem is
      (This    : in out SHFS;
       Path    : Pathname;
       Mode    : File_Mode;
-      Handler : out File_Handle_Ref)
+      Handler : out Any_File_Handle)
       return Status_Kind
    is
       FH : SHFS_File_Handle_Access;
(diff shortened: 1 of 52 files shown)


Fabien-Chouteau:
IMO, it would be OK for a class type:
`subtype Any_I2C_Port is I2C_Port'Class;`
But for an access type we lose some information and I don't like to "hide" that something is an access.

pat-rogers:
In practice, though, to be used in this way ("any" of the types in a class) it must be an access type, so there really is nothing lost.   All the more so when the naming convention is known.
(I've never seen a subtype declared like that, since 1995.)

pat-rogers:
I feel strongly that the current naming scheme is inadequate and slightly misleading, and have used the proposed scheme for many years without difficulties.

Fabien-Chouteau:
OK. I don't agree but if you feel strongly about it I'm willing to try :)

pat-rogers:
On 12/22/2016 3:54 AM, Fabien Chouteau wrote:
> OK. I don't agree but if you feel strongly about it I'm willing to try :)

Thanks.  I appreciate that.  :-)

PR #102, opened 2016-12-18
Sdcard

lambourg:
Ok, branch is now fixed, so re-sending.

The new implementation now fully supports sdcard from 1.0 to UHS-I, as well as MMC cards.

The services layer has been adapted to the change in HAL.Block_Drivers.


boards/stm32f769_discovery/board.gpr
```diff
@@ -1,7 +1,6 @@
 with "config";
 with "../../ARM/STM32/stm32f7x9";
 with "../../components/components";
-with "../../services/services";
```
Fabien-Chouteau: Why do you remove the service layer for this board?

hal/src/hal-sdcard.adb
```diff
@@ -0,0 +1,382 @@
+with Ada.Text_IO; use Ada.Text_IO;
```
Fabien-Chouteau: This is not an abstraction, it's an implementation. The HAL package should only define ways to communicate with the sdcard, the rest should be in the service layer.

BTW, is SDCard the best name for this? It's more SDC/MMC protocol.

hal/src/hal-sdcard.adb
```diff
@@ -0,0 +1,382 @@
...
+         return;
+      end if;
+
+      Send_Cmd (Driver, Go_Idle_State, 0, Status);
+      if Status /= OK then
+         return;
+      end if;
+
+      --  CMD8: IF_Cond, voltage supplied: 0x1 (2.7V - 3.6V)
+      Put_Line ("cmd: if_cond");
```
Fabien-Chouteau: Debug messages shouldn't stay here. At least they must be disabled by default.

hal/src/hal-sdcard.adb
```diff
@@ -144,12 +151,29 @@ package body HAL.SDCard is
...
+            end if;
+            return;
+         end if;
+
+         --  ACMD41: SD_SEND_OP_COND (no crc check)
+         --  Arg: HCS=1, XPC=0, S18R=0
+         Send_Cmd
+           (Driver, Acmd_Desc (SD_App_Send_Op_Cond), 16#40ff_0000#, Status);
          if Status /= OK then
             Put_Line ("send_op_cond failed");
```
Fabien-Chouteau: Debug messages shouldn't stay here. At least they must be disabled by default.

(Commenting on a line of code that is not removed by a following patch)

hal/src/hal-sdcard.adb
lambourg: right, maybe simplify it to sd.

On the services side, it's an abstract implementation of the initialisation protocol defined in the standard. I'm not sure it's service as it's support for drivers implementation (that would be a mix of layers: the service layer to me relies on drivers, not really the other way around)

boards/stm32f769_discovery/board.gpr
lambourg: Now fixed

ARM/STM32/stm32f7x9.gpr
```diff
@@ -19,6 +19,7 @@ library project STM32F7x9 is
       "drivers/dma_stm32f769",
       "drivers/dma2d",
       "drivers/dsi",
+      "drivers/jpeg",
```
lambourg: Also to be fixed

hal/src/hal-sdcard.adb
lambourg: I can confirm this can't be moved to the services layer: the drivers layer that relies on this initialisation protocol don't see the services library (and if you add it to the device project, then you end up with a circular elaboration issue).

hal/src/hal-sdcard.adb
lambourg: This is now fixed.

boards/stm32f769_discovery/src/sdcard.adb
```diff
@@ -223,8 +232,13 @@ package body SDCard is
          return;
       end if;
 
+      Enable_Clock (SDMMC_Controller (Controller.Device.all));
+      Reset (SDMMC_Controller (Controller.Device.all));
+
       Ret := STM32.SDMMC.Initialize
-        (Controller.Device.all, Controller.Info);
+        (Controller.Device.all,
+         System.BB.Board_Parameters.Main_Clock_Frequency,
```
Fabien-Chouteau: You should use STM32.Device.System_Clock_Frequencies.SYSCLK. That's what we use all across the library and it will remove the run-time dependency with the warnings off pragma.

boards/stm32f469_discovery/src/sdcard.adb
```diff
@@ -215,7 +224,9 @@ package body SDCard is
       end if;
 
       Ret := STM32.SDMMC.Initialize
-        (Controller.Device.all, Controller.Info);
+        (Controller.Device.all,
+         System.BB.Board_Parameters.Main_Clock_Frequency,
```
Fabien-Chouteau: You should use STM32.Device.System_Clock_Frequencies.SYSCLK. That's what we use all across the library and it will remove the run-time dependency with the warnings off pragma.

examples/MicroBit/src/beacon.adb
```diff
@@ -112,7 +112,7 @@ package body Beacon is
       end if;
 
       --  Set TX packet address
-      Set_Packet (Get_Address (Beacon_Packet));
+      Set_Packet (Memory_Address (Beacon_Packet));
```
Fabien-Chouteau: This shouldn't be here, you probably need a rebase.

lambourg: hmmm, indeed. Will fix.

lambourg: Ah, I see, you merged it while I was rebasing the whole stuff, that's how this ended up into this pull request.

hal/src/hal-sdcard.adb
Fabien-Chouteau: Why is the driver layer relying on the protocol layer?
How different is SDMMC from I2C or SPI? (Sending command, receiving/sending data)

ARM/STM32/svd/stm32f40x/stm32_svd-adc.ads
```diff
@@ -718,45 +719,45 @@ package STM32_SVD.ADC is
    --  Analog-to-digital converter
    type ADC1_Peripheral is record
       --  status register
-      SR    : SR_Register;
+      SR    : aliased SR_Register;
```
Fabien-Chouteau: What is the rational for this?

Also, the current version of SVD2Ada doesn't generate the output file. You must have use a custom SVD2Ada.

hal/src/hal-sdcard.ads
```diff
@@ -0,0 +1,426 @@
+with System;
+with Interfaces; use Interfaces;
+
+package HAL.SDCard is
```
Fabien-Chouteau: It should probably be HAL.SDMMC.

hal/src/hal-sdcard.ads
```diff
@@ -0,0 +1,426 @@
...
+   --  Cannot fail.
+
+   procedure Read_Rsp136
+     (This           : in out SDCard_Driver;
+      W0, W1, W2, W3 : out Unsigned_32) is abstract;
+   --  Read the 128 interesting bits of the last 136 bit response.
+   --  W0 is the MSB, W3 the LSB
+   --  Cannot fail.
+
+   procedure Card_Identification_Process (Driver : in out SDCard_Driver'Class;
```
Fabien-Chouteau: So I had a look at the code. Card_Identification_Process, Read_SCR and Swap32 are helper sub-programs that should be move to dedicated packages in the service layer. This way you won't have implementation in the HAL.

hal/src/hal-sdcard.adb
```diff
@@ -0,0 +1,661 @@
...
+            Status := OK;
+            exit;
+         end if;
+      end loop;
+
+      if Status = Command_Timeout_Error
+        and then Info.Card_Type = Multimedia_Card
+      then
+         for I in 1 .. 5 loop
+            delay until Clock + Milliseconds (200);
```
Fabien-Chouteau: You cannot use Ada.Real_Time in HAL or services, it will break ZFP compatibility.

ARM/STM32/svd/stm32f40x/stm32_svd-adc.ads
lambourg: It's the wavefront branch of SVD2Ada, waiting for the next GPL to become the master branch.

The rationale is that sometimes you need to pass register around. Having aliased values here helps.

hal/src/hal-sdcard.ads
lambourg: why not.

hal/src/hal-sdcard.ads
lambourg: I don't see the point here: this would make a mess with the current clean layers (e.g. hal, then service, then drivers, then service again). Gprbuild will certainly rightfully complain here about projects dependency circularity.

lambourg: Moreover we already have code in the HAL. I don't see why the HAL should never have code: if this code still abstracts the hardware behaviour, to me that OK. In particular if the hardware is supposed to behave in a standard manner (such is the case in SD/MMC standards, or with bluetooth for example). It would be limitating to prevent such common standard implementation there, or we should create a different layer for that, tightly close to the HAL. The service layer to me is very different, as it's above the drivers: it's pure software using services from the drivers to proceed. Filesystem is a good example, or TCP/IP.

hal/src/hal-sdcard.adb
lambourg: good catch, I forgot this one. How come it passes the tests with the new restrictions ?

lambourg: Ah, I see, the restrictions only apply to the components.

hal/src/hal-sdcard.ads
Fabien-Chouteau: > Gprbuild will certainly rightfully complain here about projects dependency circularity.

There's no circular dependency, services do not depend on drivers thanks to the HAL.

> Moreover we already have code in the HAL. 

Only in Bitmap, I plan to fix this someday.

> I don't see why the HAL should never have code: if this code still abstracts the hardware behaviour, to me that's OK.

These subprograms do not describe an interface or an abstraction, they use an interface/abstraction to compute data. They use the HAL to provide a service: the card identification.

> It would be limitating to prevent such common standard implementation there, or we should create a different layer for that, tightly linked to the HAL. 

I don't see any limitation so far, do you have an example?

> The service layer to me is very different, as it's above the drivers: it's pure software using services from the drivers to proceed. Filesystem is a good example, or TCP/IP.

What I suggest is in line with that.

HAL is at the heart of the library, it's seen by everything, it's used everywhere. The more code you put in there the more dependency you will have to deal with, that's when you will get in trouble. That why it's important to keep it as light weight and stand alone as possible.

In this specific instance there's no reason to cumber the HAL. As I said these subprograms do not describe an interface, they use an interface to provide a service.

hal/src/hal-sdcard.adb
lambourg: Well, as already told live, but for the record:

I2C or SPI protocols are 100% handled by the hardware. Here we're in a situation a bit similar to Bluetooth, where the software is supposed to initialise the sd/mmc card according to a certain standardised protocol, just so that the handshake between the controller on the card and the host can discuss properly (and efficiently). I2C or SPI don't need that.

hal/src/hal-sdcard.adb
lambourg: Fixed

boards/stm32f769_discovery/src/sdcard.adb
lambourg: Done

examples/MicroBit/src/beacon.adb
lambourg: Removed

hal/src/hal-sdcard.ads
lambourg: Now done in the sdcard branch

hal/src/hal-sdcard.ads
lambourg: For the record, I'm still not convinced.

hal/src/hal-sdcard.adb
lambourg: This is now fixed in the sdcard branch

hal/src/hal-sdcard.ads
Fabien-Chouteau: Why?

lambourg: As I said, having in the same package the low-level sd protocol, and the filesystem is just wrong, when the actual driver (another different layer) lies in the middle in a different library

Fabien-Chouteau: So you are saying that Implementation of the SDMMC software protocol should not be in the same library as file systems. Fine, if you want to create a yet another library project to put the just SDMMC software protocol that's not a huge deal. I don't think it's necessary. My point is that this code should not be in the HAL because it's not an abstraction it's an implementation.

I did the separation for you in #116, there's no circular dependency and I also put back the Cortex-M0 support that was removed for no reason.

PR #104, opened 2016-12-20
Add Travis-CI continuous integration

Fabien-Chouteau:


Fabien-Chouteau:
I forgot to mention that this is Linux 64 bits only because Travis doesn't support 32bit binaries. This means we cannot use the arm-elf-linux32 compiler.

PR #116, opened 2017-01-20
Sdcard alt

Fabien-Chouteau:
Same as the sdcard branc except:
 - Bring back Cortex-M0 sources
 - Move the software protocol part of SDMMC in the Service layer (the remaining sub-programs in hal-sdmmc.adb are no brainier utilities)


lambourg:
Well, this compiles as a library, but I still have a strong guts feeling that this won't work with Ada elaboration: 

hal < service < driver < service < application

with service being 1st low level sdcard init, then service being FS support, when this is instantiated with both a sdcard driver and an actual FS library.

Sorry I couldn't find time yet to try it out with the FAT library, I'll comment further when I do.

lambourg:
I'll try to continue on this path however. Maybe with a generic in the services layer? What do you think? This would avoid having pieces in so various places (part in HAL, part in Services, part in the Drivers, and finally part in the board itself). This leads to maintenance burden imho.

Fabien-Chouteau:
> Maybe with a generic in the services layer? What do you think?

I tried generics before, I don't know if it's logged here (maybe it was during the pre-GitHub part of this project). It's not a good feature to implement interface and abstractions, you end up propagating the instantiation over and over, it's very impractical in the end. So we will continue with the OOP way until it's proven to be a dead end or until a better solution is provided.

> This would avoid having pieces in so various places (part in HAL, part in Services, part in the Drivers, and finally part in the board itself). This leads to maintenance burden imho.

First you make it sound worst than it is, second you won't avoid splitting implementation in interface, vendor specific part and shared part. That's what we have right now and it's pretty good (HAL, drivers, services).

Having to deal with implementations in spread in different projects that would be a nightmare.

lambourg:
For the record, I'm closing this pull request as the other branch has been approved and merged.

PR #118, opened 2017-01-22
[RFC] Tentative solution for the delay problem

Fabien-Chouteau:
The idea here is that some piece of code is holds an access to an object
(Delay_Provider) that provides delay for the entire library.

The main drawback of course is that someone has to set the
Delay_Provider, in this patch I do it in the elaboration of the board
packages.

I'm not very convince myself about this solution but I though I'd try
and share it.

### DIFF
--- a/boards/OpenMV2/src/openmv-lcd_shield.adb
+++ b/boards/OpenMV2/src/openmv-lcd_shield.adb
@@ -33,7 +33,6 @@ with STM32.GPIO;
 with STM32.Device;
 with OpenMV;
 with ST7735R; use ST7735R;
-with Ravenscar_Time;
 
 package body OpenMV.LCD_Shield is
 
@@ -45,8 +44,7 @@ package body OpenMV.LCD_Shield is
    LCD_Driver : ST7735R.ST7735R_Device (Shield_SPI'Access,
                                         LCD_CS'Access,
                                         LCD_RS'Access,
-                                        LCD_RST'Access,
-                                        Ravenscar_Time.Delays);
+                                        LCD_RST'Access);
    Is_Initialized : Boolean := False;
 
    -----------------
--- a/boards/OpenMV2/src/openmv.adb
+++ b/boards/OpenMV2/src/openmv.adb
@@ -30,6 +30,8 @@
 ------------------------------------------------------------------------------
 
 with HAL.SPI;
+with HAL.Time;
+with Ravenscar_Time;
 
 package body OpenMV is
 
@@ -175,4 +177,6 @@ package body OpenMV is
    function Get_Shield_USART return not null HAL.UART.Any_UART_Port is
       (USART_3'Access);
 
+begin
+   HAL.Time.Set_Global_Delay_Provider (Ravenscar_Time.Delay_Provider);
 end OpenMV;
--- a/boards/crazyflie/src/stm32-board.adb
+++ b/boards/crazyflie/src/stm32-board.adb
@@ -41,6 +41,9 @@
 --   COPYRIGHT(c) 2014 STMicroelectronics                                   --
 ------------------------------------------------------------------------------
 
+with HAL.Time;
+with Ravenscar_Time;
+
 package body STM32.Board is
 
    -------------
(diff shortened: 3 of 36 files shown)


pat-rogers:
On 1/22/2017 4:36 AM, Fabien Chouteau wrote:
> The idea here is that some piece of code is holds an access to an object
> (Delay_Provider) that provides delay for the entire library.
>
> The main drawback of course is that someone has to set the
> Delay_Provider, in this patch I do it in the elaboration of the board
> packages.
>
> I'm not very convince myself about this solution but I though I'd try
> and share it.

Removing the issue from the client API is definitely very important and 
worthwhile, but I think you are right to be skeptical of this specific 
approach. It feels more complicated and "heavier" than necessary.

In particular, since the issue is driven ultimately by what runtime 
library is selected, I think a solution based on that selection can make 
sense. So I still think it is easiest and cleanest to have a single 
common package spec with multiple package bodies, each body reflecting 
the runtime library in place. Then we can use the RTS scenario variable 
to select the package body matching that kind of runtime. This approach 
would not address the fact that the ZFP runtimes don't have support for 
the obvious implementation using Ada.Real_Time. But then neither does 
any other approach suggested so far.

I didn't actually implement the above and propose it in a pull request 
because it is a lot of work and I wanted to discuss it first, but I am 
willing to do it.

pat-rogers:
Whether we use an existing scenario var or some other mechanism, the 
overall approach of using a specific package body for a common spec 
seems right to me.  We would need to address how to select a board- or 
family-specific package body, since they will not be as shared as the 
ravenscar-* runtimes, but I think it is doable. The trick is to share 
the body as much as possible, but not how to provide it at all. At worst 
each app project would have to provide it explicitly in their sources.

Fabien-Chouteau:
Maybe use Ada/Ravenscar when possible (native/full/sfp) and rely on something like presented in this patch for zfp.

pat-rogers:
With the DWT support I have a delay facility that can be used with ZFP and is defined by ARM so is not vendor-specific.  It samples the cycle counter in the DWT to busy-wait until the required number of cycles has passed. So I think we should have a shared package spec and two versions of the package body, one for ZFP and one for the Ravenscar-* runtimes. The package body would be selected based on the run-time, perhaps by a scenario var but maybe lower level than that.

Fabien-Chouteau:
> With the DWT support I have a delay facility that can be used with ZFP and is defined by ARM so is not vendor-specific. It samples the cycle counter in the DWT to busy-wait until the required number of cycles has passed. 

Interesting. Can we use a "Wait For Interrupt" instruction to put the CPU in sleep?

> So I think we should have a shared package spec and two versions of the package body, one for ZFP and one for the Ravenscar-* runtimes. 

Another question is where to put that package? Note that we will also have to provide one for the RISC-V as well.

> The package body would be selected based on the run-time, perhaps by a scenario var but maybe lower level than that.

We already have a variable to run-time profile so we can use that. The problem is that it's not possible to fix a range of values for a scenario variable. So for instance on the MicroBit, users will get an somewhat obscure compile error if they compile with the wrong RTS profile value.

pat-rogers:
On 3/3/2017 12:11 PM, Fabien Chouteau wrote:
>     With the DWT support I have a delay facility that can be used with
>     ZFP and is defined by ARM so is not vendor-specific. It samples the
>     cycle counter in the DWT to busy-wait until the required number of
>     cycles has passed.
>
> Interesting. Can we use a "Wait For Interrupt" instruction to put the
> CPU in sleep?

I don't think so (but could be wrong), because the loop successively 
samples the cycle counter to compare it to the target count representing 
the elapsed time requested.

>
>     So I think we should have a shared package spec and two versions of
>     the package body, one for ZFP and one for the Ravenscar-* runtimes.
>
> Another question is where to put that package? Note that we will also
> have to provide one for the RISC-V as well.

I'm not sure where to put them.  TBD.

>     The package body would be selected based on the run-time, perhaps by
>     a scenario var but maybe lower level than that.
>
> We already have a variable to run-time profile so we can use that. The
> problem is that it's not possible to fix a range of values for a
> scenario variable. So for instance on the MicroBit, users will get an
> somewhat obscure compile error if they compile with the wrong RTS
> profile value.

Mumble...

PR #120, opened 2017-01-27
Docs: Typo in design.md

Fabien-Chouteau:


Fabien-Chouteau:
> Shouldn't "Component" be plural, ie "the Components library"?

You're right, that's how the directory and project file are named.

PR #123, opened 2017-02-01
Stmpe1600

lambourg:
This pull request adds support for STMPE1600, and also changes the I2C_Data type definition to make it a subtype of Byte_Array.

### DIFF
--- a/components/src/io_expander/stmpe1600/stmpe1600.adb
+++ b/components/src/io_expander/stmpe1600/stmpe1600.adb
@@ -1,6 +1,6 @@
 ------------------------------------------------------------------------------
 --                                                                          --
---                    Copyright (C) 2017, AdaCore                           --
+--                     Copyright (C) 2017, AdaCore                          --
 --                                                                          --
 --  Redistribution and use in source and binary forms, with or without      --
 --  modification, are permitted provided that the following conditions are  --
@@ -11,7 +11,7 @@
 --        notice, this list of conditions and the following disclaimer in   --
 --        the documentation and/or other materials provided with the        --
 --        distribution.                                                     --
---     3. Neither the name of STMicroelectronics nor the names of its       --
+--     3. Neither the name of the copyright holder nor the names of its     --
 --        contributors may be used to endorse or promote products derived   --
 --        from this software without specific prior written permission.     --
 --                                                                          --
@@ -86,11 +86,11 @@ package body STMPE1600 is
       Status := S = HAL.I2C.Ok;
    end Write;
 
-   ---------------
-   -- Configure --
-   ---------------
+   --------------
+   -- Check_Id --
+   --------------
 
-   procedure Configure
+   procedure Check_Id
      (This   : in out STMPE1600_Expander;
       Status : out Boolean)
    is
@@ -103,7 +103,7 @@ package body STMPE1600 is
       then
          Status := False;
       end if;
-   end Configure;
+   end Check_Id;
 
    --------------------------
    -- Set_Interrupt_Enable --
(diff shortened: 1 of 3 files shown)


Fabien-Chouteau:
> I would comment about functions named Get_* ... :-)

Yes, we're removing them. Please do not introduce new ones.

components/src/io_expander/stmpe1600/stmpe1600.adb
```diff
@@ -0,0 +1,338 @@
...
+--  Redistribution and use in source and binary forms, with or without      --
+--  modification, are permitted provided that the following conditions are  --
+--  met:                                                                    --
+--     1. Redistributions of source code must retain the above copyright    --
+--        notice, this list of conditions and the following disclaimer.     --
+--     2. Redistributions in binary form must reproduce the above copyright --
+--        notice, this list of conditions and the following disclaimer in   --
+--        the documentation and/or other materials provided with the        --
+--        distribution.                                                     --
+--     3. Neither the name of STMicroelectronics nor the names of its       --
```
Fabien-Chouteau: If this is not derived from an STM file, please use the header like in https://github.com/AdaCore/Ada_Drivers_Library/blob/master/components/src/io_expander/MCP23xxx/mcp23x08.ads.

lambourg:
functions are now renamed

components/src/io_expander/stmpe1600/stmpe1600.adb
```diff
@@ -0,0 +1,338 @@
...
+         S);
+
+      Status := S = HAL.I2C.Ok;
+   end Write;
+
+   ---------------
+   -- Configure --
+   ---------------
+
+   procedure Configure
```
Fabien-Chouteau: The name configure seems misleading, this is only checking the device ID.

components/src/io_expander/stmpe1600/stmpe1600.ads
```diff
@@ -0,0 +1,178 @@
...
+      Pin       : STMPE1600_Pin_Number;
+      Direction : Pin_Direction);
+
+   procedure Set_Pin_Polarity_Inversion
+     (This            : STMPE1600_Expander;
+      Pin             : STMPE1600_Pin_Number;
+      Inversion_State : Boolean);
+   --  Enables or disables the polarity inversion of the given input pin
+
+   type STMPE1600_Pin
```
Fabien-Chouteau: This type doesn't have to be public. See https://github.com/AdaCore/Ada_Drivers_Library/blob/master/components/src/io_expander/MCP23xxx/mcp23x08.ads

lambourg: I don't understand why making this type private? This makes the use of the HAL.GPIO interface useless (e.g. the ability to pass pins to a specific component, without having the explicit board definition).

In my case, this would break the implementation of the VL53L0x driver for example, as it needs access to the reset pin of the sensor, and I don't want to have it depending on a specific gpio expander.

components/src/io_expander/stmpe1600/stmpe1600.adb
lambourg: Renamed to Check_Id

components/src/io_expander/stmpe1600/stmpe1600.adb
lambourg: Done

components/src/io_expander/stmpe1600/stmpe1600.ads
Fabien-Chouteau: > I don't understand why making this type private? This makes the use of the HAL.GPIO interface useless (e.g. the ability to pass pins to a specific component, without having the explicit board definition).
 

As I said in my comment, please have a look at the MCP23x08. There's a function that returns the HAL GPIO_Point corresponding to a given pin on the expender. It's much simpler for the users and the type stays private.

> In my case, this would break the implementation of the VL53L0x driver for example, as it needs access to the reset pin of the sensor, and I don't want to have it depending on a specific gpio expander.

I would never suggest something like that.

lambourg: Ok, I missed the function when I first had a look.

This is now implemented.

PR #124, opened 2017-02-01
Correct scan mode enabling in configure routines

pat-rogers:
Also ensures consistent external trigger and enabler in subsequent calls; minor simplifications


Fabien-Chouteau:
There's something strange with the pull request. Can you try to do a pull --rebase and then push the branch again?

pat-rogers:
On 2/1/2017 10:06 AM, Fabien Chouteau wrote:
> There's something strange with the pull request. Can you try to do a pull --rebase and then push the branch again?

OK, just did  and pushed (a new branch).

Fabien-Chouteau:
These "merge" commits probably come when you do a pull without the --rebase option. I don't think it will cause a problem but if we can get rid of them that would be better.

pat-rogers:
On 2/1/2017 10:21 AM, Fabien Chouteau wrote:
> These "merge" commits probably come when you do a pull without the --rebase option. I don't think it will cause a problem but if we can get rid of them that would be better.

I'm using TortoiseGit to do most things, including the pull (because it 
is convenient) and can enable the "Rebase after fetch" option when doing so.

Fabien-Chouteau:
It looks like it doesn't work how it's supposed to.

For the next time, can you try to add this to your config file:

`[branch]
  autosetuprebase = always`

or
`$ git config --global branch.autosetuprebase always`

pat-rogers:
On 2/1/2017 11:17 AM, Fabien Chouteau wrote:
> It looks like it doesn't work how it's supposed to.
>
> For the next time, can you try to add this to your config file:
>
> `[branch]
>   autosetuprebase = always`
>
> or
> `$ git config --global branch.autosetuprebase always`

Done (the latter).

PR #127, opened 2017-02-02
Change enum value name to better match the F4 series Ref Manual

pat-rogers:


pat-rogers:
Since this one is so trivial I will merge it now, rather than give it the usual waiting period...

PR #128, opened 2017-02-02
GPIO: mode and pull control

Fabien-Chouteau:
Add primitives to control the mode (input/output) and pull resistors of GPIO points.


Fabien-Chouteau:
> Seems like the name GPIO_Pull should have the term "Resistor" in there somewhere, to reflect the domain terminology.

I agree and I updated fixed it.

PR #133, opened 2017-02-07
Update README.md

evilspacepirate:
Add component: BlueNRG-MS

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,7 @@ GitHub page.
 * ARM
  * STM32F4/7 USB drivers
 * Components
+ * BlueNRG-MS (Bluetooth Low Energy Network Processor)
 * Services
  * Bluetooth Low Energy stack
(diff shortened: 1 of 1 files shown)


Fabien-Chouteau:
Hi,
Do you plan to implement an Ada firmware for this device or a driver to control it from outside?

I'm also curious to know in which project your are going to use this chip :)

evilspacepirate:
Fabien,

The goal is to control the BlueNRG-MS component from the outside using an
STM32F407 chip.

John

Fabien-Chouteau:
> The goal is to control the BlueNRG-MS component from the outside using an STM32F407 chip.

Thanks, I will apply your patch. Don't hesitate to give us progress report or to do intermediate pull-requests to get some early feedback.

evilspacepirate:
Thanks Fabien.

PR #135, opened 2017-02-08
correct descriptor comment for CRC units

pat-rogers:
Minor fix to the comment lines indicating the peripheral. These comments match those committed elsewhere to the SVD files so this is what would be generated.

### DIFF
--- a/ARM/STM32/svd/stm32f40x/stm32_svd-crc.ads
+++ b/ARM/STM32/svd/stm32f40x/stm32_svd-crc.ads
@@ -49,10 +49,10 @@ package STM32_SVD.CRC is
    -- Peripherals --
    -----------------
 
-   --  Cryptographic processor
+   --  Cyclic Redundancy Check (CRC) unit
    type CRC_Peripheral is record
       --  Data register
-      DR  : HAL.UInt32;
+      DR  : HAL.UInt32 with Volatile_Full_Access;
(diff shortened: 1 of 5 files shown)


ARM/STM32/svd/stm32f40x/stm32_svd-crc.ads
```diff
@@ -49,10 +49,10 @@ package STM32_SVD.CRC is
    -- Peripherals --
    -----------------
 
-   --  Cryptographic processor
+   --  Cyclic Redundancy Check (CRC) unit
    type CRC_Peripheral is record
       --  Data register
-      DR  : HAL.UInt32;
+      DR  : HAL.UInt32 with Volatile_Full_Access;
```
Fabien-Chouteau: You left one Volatile_Full_Access here. Is that on purpose?

pat-rogers:
On 2/8/2017 10:59 AM, Fabien Chouteau wrote:
> Fabien-Chouteau commented on this pull request.
>
>
>
>>     type CRC_Peripheral is record
>        --  Data register
> -      DR  : HAL.UInt32;
> +      DR  : HAL.UInt32 with Volatile_Full_Access;
>
> You left one Volatile_Full_Access here. Is that on purpose?

Oops, no. I'll fix it.

PR #139, opened 2017-02-10
Add the CRC driver and demo for the F4 series.

pat-rogers:
Note that the F3 and F7 have more flexibility and thus a different API.
For the SVD-generated packages, add the Volatile aspect as a temporary work-around for a compiler bug.


ARM/STM32/drivers/crc_stm32f4/stm32-crc.ads
```diff
@@ -0,0 +1,107 @@
...
+     with Component_Size => 32;
+
+   procedure Update_CRC
+     (This   : in out CRC_32;
+      Input  : Block_32;
+      Output : out UInt32);
+   --  Updates the 32-bit CRC value in This from the 32-bit Input values.
+   --  Output is the resulting CRC-32 value.
+
+   type Block_16 is array (Positive range <>) of UInt16
```
Fabien-Chouteau: We already have a UInt16_Array type in HAL, maybe you could use it here.

ARM/STM32/drivers/crc_stm32f4/stm32-crc.ads
```diff
@@ -0,0 +1,107 @@
...
+     with Component_Size => 16;
+
+   procedure Update_CRC
+     (This   : in out CRC_32;
+      Input  : Block_16;
+      Output : out UInt32);
+   --  Updates the 32-bit CRC value in This from the 16-bit Input values.
+   --  Output is the resulting CRC-32 value.
+
+   type Block_8 is array (Positive range <>) of UInt8
```
Fabien-Chouteau: Same here.

pat-rogers:
On 2/10/2017 9:49 AM, Fabien Chouteau wrote:
> Fabien-Chouteau commented on this pull request.
>
> Only two minor comments.
>
>> +      Input  : UInt32;
> +      Output : out UInt32);
> +   --  Updates the 32-bit CRC value in This from the 32-bit Input value.
> +   --  Output is the resulting CRC-32 value.
> +
> +   type Block_32 is array (Positive range <>) of UInt32
> +     with Component_Size => 32;
> +
> +   procedure Update_CRC
> +     (This   : in out CRC_32;
> +      Input  : Block_32;
> +      Output : out UInt32);
> +   --  Updates the 32-bit CRC value in This from the 32-bit Input values.
> +   --  Output is the resulting CRC-32 value.
> +
> +   type Block_16 is array (Positive range <>) of UInt16
>
> We already have a UInt16_Array type in HAL, maybe you could use it here.
>
>> +      Input  : Block_32;
> +      Output : out UInt32);
> +   --  Updates the 32-bit CRC value in This from the 32-bit Input values.
> +   --  Output is the resulting CRC-32 value.
> +
> +   type Block_16 is array (Positive range <>) of UInt16
> +     with Component_Size => 16;
> +
> +   procedure Update_CRC
> +     (This   : in out CRC_32;
> +      Input  : Block_16;
> +      Output : out UInt32);
> +   --  Updates the 32-bit CRC value in This from the 16-bit Input values.
> +   --  Output is the resulting CRC-32 value.
> +
> +   type Block_8 is array (Positive range <>) of UInt8
>
> Same here.


A toss-up, but I like to keep a spec's declarations as local as 
possible, unless there is a compelling reason not to.

(By the way, those in the HAL should have their component size aspect 
specified, just to be 100% sure there are no gaps. There wouldn't be 
gaps with those component types in any reasonable implementation, but it 
wouldn't hurt.)

Fabien-Chouteau:
> A toss-up, but I like to keep a spec's declarations as local as possible, unless there is a compelling reason not to. 

Being able to use the same array for CRC as for SPI, I2C coms.

pat-rogers:
On 2/10/2017 11:09 AM, Fabien Chouteau wrote:
>> A toss-up, but I like to keep a spec's declarations as local as possible, unless there is a compelling reason not to.
>
> Being able to use the same array for CRC as for SPI, I2C coms.

Do you mean there is a use-case for calculating the CRC value for arrays 
used in SPI and I2C usage, or do you mean the convenience of using the 
same array type as for those protocols?

(The array types are convertible, in any case.)

PR #141, opened 2017-02-11
Start experimental support of the FE310 RISC-V MCU

Fabien-Chouteau:
This PR introduces a new architecture and move the existing ARM support in the arch subdir.

As said in the README, the support is experimental.


Fabien-Chouteau:
> We should address the package Interfaces issue sooner than later.

You mean the HAL types that derives from Interfaces types?

pat-rogers:
On 2/13/2017 11:27 AM, Fabien Chouteau wrote:
>> We should address the package Interfaces issue sooner than later.
>
> You mean the HAL types that derives from Interfaces types?

Yes, but currently they aren't derived -- they're just subtypes -- and 
that's the primary source of the issue.

The issue is that client code often needs a with/use-clause for 
Interfaces to get the operators for the subtypes declared in HAL, in 
addition to a clause for HAL, because subtypes just introduce names for 
constraints. Subtypes do not make the operators visible at the point of 
the subtype declaration, whereas derived types do make the operators 
visible.

So we should say:

    type UInt32 is new Interfaces.Unsigned_32;

instead of

    subtype UInt32 is Interfaces.Unsigned_32;

and so on for the others declared in Interfaces.

That would make the shift and rotate functions available too. Note that 
user-defined modular types do not have those functions defined. That's a 
separate issue though.

The other issue is that some of the code uses the signed integer types 
in Interfaces too, so making the subtypes into derived types won't 
address that use of Interfaces. Maybe we should declared signed types in 
HAL?

Fabien-Chouteau:
I will try to do that (not in this PR) but it's a big change. 

For the signed types I don't know...

pat-rogers:
On 2/13/2017 12:15 PM, Fabien Chouteau wrote:
> I will try to do that (not in this PR) but it's a big change.

Much of it will just be the initial change to the HAL package, and then 
removal of a bunch of with-use clauses for Interfaces (maybe with 
replacement by with/use clauses for HAL).  Followed by textual 
replacement of Unsigned_x with UIntx in a lot of places.

> For the signed types I don't know...

Looks like there are around 24 places where they are used.

Mostly this is pretty straightforward changes, but the testing is 
another matter.

But the sooner we do it the less there will be to do, compared to 
propagating the issue.

I'll sign up to do it if you want.

PR #143, opened 2017-02-11
Add BlueNRG-MS driver for the STM32F4

evilspacepirate:
The BlueNRG-MS is a Bluetooth Low Energy (BLE) Network Processor supporting Bluetooth 4.1 core specification.

### DIFF
--- a/components/src/network/bluenrg-ms/bluenrg_ms-debug.adb
+++ b/components/src/network/bluenrg-ms/bluenrg_ms-debug.adb
@@ -1,0 +1,244 @@
+----------------------------------------------------------------------------
+--                                                                        --
+--  BlueNRG-MS.Debug                                                      --
+--                                                                        --
+--  Copyright (C) 2017, John Leimon                                       --
+--                                                                        --
+-- Permission to use, copy, modify, and/or distribute                     --
+-- this software for any purpose with or without fee                      --
+-- is hereby granted, provided that the above copyright                   --
+-- notice and this permission notice appear in all copies.                --
+--                                                                        --
+-- THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR                        --
+-- DISCLAIMS ALL WARRANTIES WITH REGARD TO THIS SOFTWARE                  --
+-- INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY                    --
+-- AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE                    --
+-- FOR ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL                    --
+-- DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS                  --
+-- OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF                       --
+-- CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING                 --
+-- OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF                 --
+-- THIS SOFTWARE.                                                         --
+--                                                                        --
+----------------------------------------------------------------------------
+
+package body BlueNRG_MS.Debug is
+   
+   LF   : constant Character := Character'Val(16#0A#);
+   CR   : constant Character := Character'Val(16#0D#);
+   CRLF : constant String    := CR & LF;
+
+   -------------------------
+   -- HCI_Event_To_String --
+   -------------------------
+
+   function HCI_Event_To_String
+     (Data : Byte_Array)
+      return String
+   is
+   begin
+      if Data'Length = 0 then
+         return "No Data";
+      end if;
+
+      if Data(1) /= 16#04# then
+         return "Not an Event";
+      end if;
+
+      if Data'Length < 2 then
+         return "Not enough data";
+      end if;
+
+      if Data(2) /= 16#FF# then
+         return Bluetooth.Debug.HCI_Event_To_String(Data);
+      end if;
+
+      if Data(5) = 16#08# and
+         Data(4) = 16#00#
+      then
+         return "Evt_Blue_L2CAP_Conn_Upd_Resp";
+      end if;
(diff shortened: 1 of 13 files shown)


evilspacepirate:
Fabien,

I used the STM32F4 LED demo template from GPS as a basis for my project. I
modified the project files from there.

How would you recommend imposing delays without using Ada.Real_Time?

Thanks,
John

Fabien-Chouteau:
> I used the STM32F4 LED demo template from GPS as a basis for my project. I modified the project files from there. 

I would recommend to start from a project of Ada_Drivers_Library, "examples/accelerometer" is a good one. You first copy the directory in your project and then change the path in "accelerometer.gpr" to point to the Ada_Drivers_Library.

> How would you recommend imposing delays without using Ada.Real_Time?

The solution we have right now - which is not far from satisfactory but it works - is to use an HAL.Time.Any_Delays interface. You can see an example of that in the [ST7735R](https://github.com/AdaCore/Ada_Drivers_Library/blob/master/components/src/screen/ST7735R/st7735r.ads).

Since you are using Ravenscar, the Delays interface will be provided by the [Ravenscar_Time](https://github.com/AdaCore/Ada_Drivers_Library/blob/master/middleware/src/ravenscar-common/ravenscar_time.ads) package. Here's how to use it:

```
with Ravenscar_Time;

package body OpenMV.LCD_Shield is
   [...]
   LCD_Driver : ST7735R_RAM_Framebuffer_Device (Shield_SPI'Access,
                                                LCD_CS'Access,
                                                LCD_RS'Access,
                                                LCD_RST'Access,
                                                Ravenscar_Time.Delays) --  < Using Ravenscar delays here;
```

evilspacepirate:
>You first copy the directory in your project...

Should copy my project's source code files to the examples/accelerometer directory, or should I copy the examples/accelerometer directory outside of the Ada_Drivers_Library directory (e.g. Desktop)?

>...then change the path in "accelerometer.gpr" to point to the Ada_Drivers_Library.

Which path in accelerometer.gpr are you referring to? The only paths that look modifiable are the:

`with "../../boards/stm32f407_discovery.gpr`
and
`project Accelerometer extends "../common/common.gpr`

Fabien-Chouteau:
> Should I copy my project's source code files to the examples/accelerometer directory, or should I copy the examples/accelerometer directory outside of the Ada_Drivers_Library directory (e.g. Desktop)?
Both will work, but I would recommend copying outside of Ada_Drivers_Library so you can control what goes in which project. 

>     ...then change the path in "accelerometer.gpr" to point to the Ada_Drivers_Library.
> 
> Which path in accelerometer.gpr are you referring to? The only paths that look modifiable are:
> 
> with "../../boards/stm32f407_discovery.gpr
> and
> project Accelerometer extends "../common/common.gpr

Yes, I'm talking about those two paths. Since you are copying the project, those two relative paths have to be changed.

evilspacepirate:
I've copied the accelerometer directory to a path outside of the ada_drivers_library directory and I've modified the two lines to point to the ada_drivers_library:

`with "../ada_drivers_library/boards/stm32f407_discovery.gpr";`

`project Accelerometer extends "../ada_drivers_library/examples/common/common.gpr"`

When I try to build the modified accelerometer project with GPS, GPS displays the message:

`[2017-02-14 13:43:46] variable"rts" is not a single string variable`

When I build it with gprbuild on the command line, I get the following output:

`/Desktop/accelerometer $ gprbuild`
`using project file accelerometer.gpr`
`variable"rts" is not a single string variable`
`gprbuild: raised PROGRAM_ERROR : gpr-proc.adb:2403 explicit raise`
`Call stack traceback locations:`
`0x5873ce 0x587fdc 0x588b7a 0x587ece 0x588b7a 0x587ece 0x58799e 0x5886a5 0x588b7a 0x587ece 0x58799e 0x57cf86 0x411830 0x40c1e5 0x7f6c623c9ec3 0x40c23f`

Additionally, after loading the accelerometer.gpr in GPS the Project view does not show any source files (e.g. main.adb).

Software Versions:
gprbuild: GPRBUILD GPL 2016 (20160515) (x86_64-pc-linux-gnu)
GPS: (20160515) hosted on x86_64-pc-linux-gnu
GNAT GPL 2016 (20160515-49)

Fabien-Chouteau:
This looks like a problem in the paths. There's something shady with the run-time.

Don't get stuck on this issue and forget what I said, start by modifying the example inside the Ada_Drivers_Library if it's easier.

evilspacepirate:
Is there a method to compare two Time variables without using Ada.Real_Time? I have a case where I want to continue executing a Read operation until a certain timeout is reached.

Fabien-Chouteau:
> s there a method to compare two Time variables without using Ada.Real_Time? I have a case where I want to continue executing a Read operation until a certain timeout is reached.

I guess you are asking because this code is in the middle-ware and we forbid Ada.Real_Time. Do you get the time values from the BlueNRG or are you also looking for ways to get a time value?

evilspacepirate:
I am not receiving time from the BlueNRG device. I was looking for a way to get the time value in order to determine that a time out condition has occurred. If this is not possible I will experiment with counting delay operations.

Fabien-Chouteau:
OK, we don't have a good solution for this right now but we started to discuss a different things in #118.
The problem is that Ada.Real_Time is not available on all platforms.

In the meantime, you can add a function that will return a clock value in the [HAL.Time.Delays](https://github.com/AdaCore/Ada_Drivers_Library/blob/master/hal/src/hal-time.ads) interface and the implement that function in the [Ravenscar_Time package](https://github.com/AdaCore/Ada_Drivers_Library/blob/master/middleware/src/ravenscar-common/ravenscar_time.ads).

Or remove the Ada.Real_Time restriction in your branch so that you can make progress on your driver until we find a better solution.

simonjwright:
John,

A lot of those style warnings were about trailing blanks. You can set GPS to strip trailing blanks automatically (Preferences > Editor > Strip blanks; not sure what 'Autodetect' means, perhaps it means only do this for Ada source).

evilspacepirate:
Closing this request for now. After the code is refactored, I'll try again.

PR #146, opened 2017-02-13
HAL: Make UInt8|16|32 derived from Interfaces types rather than subtypes

Fabien-Chouteau:
As discussed in #141.

This remove the dependency on the Interfaces package which simplify the
use of HAL types.

Also remove Byte and Bit subtype.

### DIFF
--- a/arch/ARM/Nordic/drivers/nrf51-radio.adb
+++ b/arch/ARM/Nordic/drivers/nrf51-radio.adb
@@ -298,7 +298,7 @@ package body nRF51.Radio is
    ----------------------
 
    procedure Configure_Packet
-     (S0_Field_Size_In_Byte        : UInt1;
+     (S0_Field_Size_In_Byte        : Bit;
       S1_Field_Size_In_Bit         : UInt4;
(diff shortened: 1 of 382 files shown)


Fabien-Chouteau:
@pat-rogers I didn't check the STM32 drivers demo yet.

Fabien-Chouteau:
> type UInt1 is mod 2**1;
> subtype Bit is UInt1;

We can do one or the other, I don't have a strong opinion here. But I don't think it's a good idea to have both.

pat-rogers:
On 2/14/2017 4:59 AM, Fabien Chouteau wrote:
>> type UInt1 is mod 2**1;
>> subtype Bit is UInt1;
>
> We can do one or the other, I don't have a strong opinion here. But I don't think it's a good idea to have both.

OK. I think you were right to name that one Bit originally, because it 
is so much more readable (and is unambiguous, unlike byte etc.).

Fabien-Chouteau:
Anyone else want's to comment on the UInt1 vs Bit discussion? Otherwise I will revert to Bit.

PR #151, opened 2017-02-18
New "window" watchdog driver and demo

pat-rogers:


pat-rogers:
On 2/20/2017 9:08 AM, Fabien Chouteau wrote:
> Fabien-Chouteau approved this pull request.
>
> Outstanding code quality as always :)

Thanks! :-)

> Is is possible to provide a helper that will compute the prescaler based on the frequency that the users want to have?

AFAICT there is not the same flexibility as we had with the PWM 
implementation, for example.  The clock source is always PCLK1 and the 
dividers are always 1, 2, 4, and 8.

PR #152, opened 2017-02-19
 STM32F4.I2C: Fix master transmit

Fabien-Chouteau:
Looks like this code was badly copied from the reference drivers. We are
not supposed to wait for BTF to be set, but just check it once. Note
that when reading the documentation, the behavior of this BTF flag is
still not clear to me. I think I spotted this error by reducing the
speed on an I2C port for an experiment. The result was that one out of
two bytes was dropped.


lambourg:
Note that https://github.com/AdaCore/Ada_Drivers_Library/commit/15bacaf3123528095b6c0469ab202b172d534646 fixes two other issues there: one wrong flag check and a minor one on renaming of the expected flag name itself.

Fabien-Chouteau:
@pat-rogers This will be for another patch. This PR is fixing a bug in the library it needs to be applied ASAP.

PR #156, opened 2017-02-21
Use relative path in .gitmodules

orgads:
Let the user choose how to clone.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "embedded-runtimes"]
 	path = embedded-runtimes
-	url = https://github.com/AdaCore/embedded-runtimes.git
+	url = ../embedded-runtimes


Fabien-Chouteau:
Hi @orgads,

Thanks for your contribution.

I'm think I'm OK with this change. Can you tell us what to you mean by "choose how to clone"? What can you do differently with a relative path?

Thanks,

orgads:
You can clone using ssh, or even clone from a different fork, and use the corresponding submodule.

For example, if someone forks both repos, and changes both of them (including the reference), this won't work without using a relative path.

Fabien-Chouteau:
> For example, if someone forks both repos, and changes both of them (including the reference), this won't work without using a relative path.

The corollary is that you are forced to clone both repositories, right?

orgads:
>> For example, if someone forks both repos, and changes both of them (including the reference), this won't work without using a relative path.

> The corollary is that you are forced to clone both repositories, right?

Not necessarily. I usually clone from the _original_ repo, then I add my fork with `git remote add my <address>`. In that case, `git submodule update --init` and `git clone --recursive` will use the original repo.

Fabien-Chouteau:
That sounds a bit too difficult for those who are not git/GitHub expert.
I think we are going to keep the current URL unless many people request this.

Thanks,

PR #162, opened 2017-02-25
Sdcard

lambourg:
This moves the sd/mmc cards initialisation protocol to a common directory, to be used by any sd or mmc card driver.


lambourg:
I finally updated this branch to the latest API changes, and moved the SDMMC init to the middleware folder.

Note that I haven't re-introduced the notion of HAL.SDMMC as the only interface that is useful is already there, and that's the HAL.Block_Driver.

Fabien-Chouteau:
Thanks Jerome,

> Note that I haven't re-introduced the notion of HAL.SDMMC as the only interface that is useful is already there, and that's the HAL.Block_Driver.

I don't understand why SDMMC would not be a useful interface, It can be implemented by other MCU drivers it can even be implemented over a SPI interface. It can also be used to implement more than Block_Driver (Camera, GPS, Wifi, etc). That's the only blocking point for me.

Other "minor" comments:

Looks like sdcard.ad[bs] code should either be in STM32.SDMMC or in a platform agnostic Block_Driver over SDMMC interface. I don't see how this is specific to the STM32F469 discovery board. In the end, the amount of work to get a Block_Driver interface working seems quite high, even for someone using the same chip just on a different board.

Also I don't like the System.Address buffer + UInt32 Len arguments in the SDMMC interface, it's not matching what we've done so far and I'm not fond of this kind of void* buffer passing.

lambourg:
On the sdcard.ad[bs], it's the same issue as the framebuffer or audio ones that are also declared within the disco board definitions.

They share in general a lot in common, so I intend to refactor those at some point, so as to maximise the reuse there. For now it's mostly copy/paste and it's a burden to maintain.

Anyway, that'll be part of a different pull request, I wanted to concentrate for this request just on the refactor of the sd/mmc init as defined by the standard.

lambourg:
Hi Fabien,

Now on the blocking point, to me that's not an issue. Let me explain my point of view:

The sdmmc_init package that is introduced by this pull request only concerns the initialisation phase. So it's needed by the driver to perform it's init and just that (which he could do by using the middleware/sdmmc.gpr project introduced by this pull request, and by just providing a regular "Configure" subprogram to the end-user). This init phase is purely internal, and does not provide any added value to the end user. That's where I don't understand what kind of sd interface in the HAL would be useful here, and what interface would be in common between a camera and a sdcard. The only thing I could think of is the error code returned by the init. If that's what you have in mind, then why not?

Finally, I have zero experience with other peripherals using the SD protocol, so can't even say that the current SDMMC_Init package as implemented here would be useful in such situation.

Fabien-Chouteau:
> Note that I don't want to integrate them into STM32.SDMMC as they contain interrupt handlers, and as far as I know we don't want such thing in the drivers:  

The idea to give the choice, because we cannot know what is best for every possible project.

> that's for the user of the drivers to determine whether we want interrupts or not. 

Letting the users choose between polling and interrupt doesn't mean letting them implement more stuff. DMA2D and RNG are good examples.

> I know that we have an exception with the 'interrupt' variant initialisation of the DMA2D, but even that would IMHO be better out of the drivers.

Why would it be better out of the drivers? And where would you implement it?

Note that I started to work on easier DMA usage and DMA SPI and I2C variant. I hope I will find some time to make a first pull request. (in the stm32_dma_services branch)

Fabien-Chouteau:
> The sdmmc_init package that is introduced by this pull request only concerns the initialisation phase.

That's your only concern at the moment but it doesn't mean it's the only possible usage for it. In particular, I think that you do not use SDMMC_Driver interface as much as you should. For instance, could you do a MCU agnostic implementations of Block_Driver using the Read_CMD () primitive of SDMMC_Driver?

> So it's needed by the driver to perform it's init and just that (which he could do by using the middleware/sdmmc.gpr project introduced by this pull request, and by just providing a regular "Configure" subprogram to the end-user). This init phase is purely internal, and does not provide any added value to the end user.

As I said above I don't think this interface is only useful for the init phase. Also I don't think it's a good idea to speculate too much on the end user and what he/she wants.

> That's where I don't understand what kind of sd interface in the HAL would be useful here, and what interface would be in common between a camera and a sdcard.

The things that are the heart of the SDMMC protocol: sending commands, reading and writing blocks.

Then you implement the GPS, camera or block driver on top of the SDMMC interface like we implement drivers for gyroscope or screens on top I2C or SPI.

> Finally, I have zero experience with other peripherals using the SD protocol, so can't even say that the current SDMMC_Init package as implemented here would be useful in such situation.

Even if you don't implement them, exploring the different capabilities of a protocol helps to define better interfaces because you can see what part has to be shared, what is the least common denominator. It's also important to see the controllers of other vendors. For instance with SDMMC, it's possible and probably more common for MCUs to use an SPI controller to talk to the card. How would you move all your SDMMC + Block_Driver + FATfs stack to an SPI controller?

lambourg:
No, that's not my concern, that's that standard's concern. It only speaks about initialisation. Nothing else.

lambourg:
Hi Fabien,

In case you haven't looked at the latest commits, the code is now updated according to your last comments.

- Jerome

PR #164, opened 2017-02-26
Allow composition using PWM type by removing timer discriminant.

pat-rogers:
Also make the PWM type tagged, since it is not a direct driver (it is a higher-level wrapper).

The issue with composition is that, for example, an enclosing record type cannot simply declare a component of the PWM type because the discriminant (timer) constraint is required at that point, and it might not be known there. That forces the component to be an access to the PWM type, thus requiring dynamic allocation or a fix timer object, or a timer discriminant on the enclosing type -- both of which are  problematic).


Fabien-Chouteau:
> The issue with composition is that, for example, an enclosing record type cannot simply declare a component of the PWM type because the discriminant (timer) constraint is required at that point, and it might not be known there.

I think we will have the same problem with components where we use discriminants a lot.

pat-rogers:
> I think we will have the same problem with components where we use discriminants a lot.

Yes, we'll need to be careful.

PR #170, opened 2017-03-04
[RFC] STM32 easy device setup

Fabien-Chouteau:
The goal of this PR is to provide easier setup of common STM32 devices (I2C, SPI, UART, I2S). This kind of setup code is duplicated a lot in the ADL so will be able to both factorize and provide an better experience for users.

This is only an RFC right now, not all the setups are implemented and/or used.


pat-rogers:
I like the first two a lot, the third makes sense too.

PR #174, opened 2017-03-07
- update for issue with mem addr 16b parts.

morbos:
Tested with 8b BNO055 part to verify 8b not busted.

Original issue was with 16b mem addr, the read/write cmd following the write of the mem addr LSB would step on the LSB of the addr. Interestingly the MSB to LSB TxE wait is basically a nop as the part immediately reports TxE and latches the LSB, the next write to the bus though... that one needs a wait as that is where the TxE hold will wait.

### DIFF
--- a/arch/ARM/STM32/drivers/i2c_stm32f4/stm32-i2c.adb
+++ b/arch/ARM/STM32/drivers/i2c_stm32f4/stm32-i2c.adb
@@ -561,6 +561,14 @@ package body STM32.I2C is
 
             This.Periph.DR.DR := UInt8 (Mem_Addr and 16#FF#);
       end case;
+
+      --  Wait until TXE flag is set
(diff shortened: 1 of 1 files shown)


Fabien-Chouteau:
Thanks Hedley,

Can you associate the email address used in your commit with your GitHub account, this is needed for the CLA verification tool.

The procedure is described [here](https://help.github.com/articles/adding-an-email-address-to-your-github-account/).

Thanks,

morbos:
Thanks. Unfortunately, that commits email cannot ever be verified (never having collaborated before, this was not noticed till now). Future checkins now have a real email associated with them. Don't know how to retroactively change the commit to the new, verified email.

Fabien-Chouteau:
> Can you delete this pull completely and I will resubmit?

OK, let's do that :)

PR #175, opened 2017-03-09
Fix for mem addr 16b TxE wait issue

morbos:
Tested with 8b BNO055 part to verify 8b not busted.

Original issue was with 16b mem addr, the read/write cmd following the write of the mem addr LSB would step on the LSB of the addr. Interestingly the MSB to LSB TxE wait is basically a nop as the part immediately reports TxE and latches the LSB, the next write to the bus though... that one needs a wait as that is where the TxE hold will wait.

### DIFF
--- a/arch/ARM/STM32/drivers/i2c_stm32f4/stm32-i2c.adb
+++ b/arch/ARM/STM32/drivers/i2c_stm32f4/stm32-i2c.adb
@@ -561,6 +561,14 @@ package body STM32.I2C is
 
             This.Periph.DR.DR := UInt8 (Mem_Addr and 16#FF#);
       end case;
+
+      --  Wait until TXE flag is set
+      Wait_Flag (This, Tx_Data_Register_Empty, False, Timeout, Status);
+
+      if Status /= HAL.I2C.Ok then
+         return;
+      end if;
+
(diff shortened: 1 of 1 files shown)


Fabien-Chouteau:
Thanks Hedley,

I want to check with other boards before merging but I don't have time right now. Next week maybe.

arch/ARM/STM32/drivers/i2c_stm32f4/stm32-i2c.adb
```diff
@@ -561,6 +561,14 @@ package body STM32.I2C is
 
             This.Periph.DR.DR := UInt8 (Mem_Addr and 16#FF#);
       end case;
+
+      --  Wait until TXE flag is set
+      Wait_Flag (This, Tx_Data_Register_Empty, False, Timeout, Status);
+
+      if Status /= HAL.I2C.Ok then
```
Fabien-Chouteau: Hello Hedley,

Sorry for the delay. 

This if then return is not needed. Can you modify this patch or add the new commit that remove those 3 lines?

Thanks in advance,

Fabien-Chouteau:
Thanks Hedley!

PR #176, opened 2017-03-09
Add Transmit_Receive to HAL.SPI SPI_Port interface

evilspacepirate:
Add Transmit_Receive procedure to the SPI_Port interface.

Rationale:
1. Only the STM32-SPI package currently uses this interface, and it already has the Transmit_Receive procedure implemented.
2. The STMicroelectronics BlueNRG-MS component requires a transmit and receive operation to communicate with the component using the SPI interface. See STMicroelectronics UM1865 "BlueNRG-MS Bluetooth LE stack application command interface (ACI)" section 5.2.1. Figure 10 SPI frame header. In this protocol, a 'control' byte is transmitted and a 'ready' byte is received simultaneously.

### DIFF
--- a/arch/ARM/STM32/drivers/stm32-spi.adb
+++ b/arch/ARM/STM32/drivers/stm32-spi.adb
@@ -921,6 +921,7 @@ package body STM32.SPI is
    -- Transfer --
    --------------
 
+   overriding
    procedure Transfer
      (This      : in out SPI_Port;
       Data_Out  :        HAL.SPI.SPI_Data_8b;
@@ -930,8 +931,8 @@ package body STM32.SPI is
    is
       pragma Unreferenced (Timeout);
    begin
-      Send_Receive_8bit_Mode(This, UInt8_Buffer (Data_Out), UInt8_Buffer (Data_In), Data_Out'Length);
-      Status := HAL.SPI.OK;
+      Send_Receive_8bit_Mode (This, UInt8_Buffer (Data_Out), UInt8_Buffer (Data_In), Data_Out'Length);
+      Status := HAL.SPI.Ok;
    end Transfer;
 
    ------------------------
--- a/arch/ARM/STM32/drivers/stm32-spi.adb
+++ b/arch/ARM/STM32/drivers/stm32-spi.adb
@@ -917,6 +917,23 @@ package body STM32.SPI is
       end if;
    end Send_8bit_Mode;
 
+   --------------
+   -- Transfer --
+   --------------
+
+   procedure Transfer
+     (This      : in out SPI_Port;
+      Data_Out  :        HAL.SPI.SPI_Data_8b;
+      Data_In   :    out HAL.SPI.SPI_Data_8b;
+      Status    :    out HAL.SPI.SPI_Status;
+      Timeout   :        Natural := 1000)
+   is
+      pragma Unreferenced (Timeout);
+   begin
+      Send_Receive_8bit_Mode(This, UInt8_Buffer (Data_Out), UInt8_Buffer (Data_In), Data_Out'Length);
+      Status := HAL.SPI.OK;
+   end Transfer;
+
    ------------------------
    -- Receive_16bit_Mode --
    ------------------------
--- a/arch/ARM/STM32/drivers/stm32-spi.ads
+++ b/arch/ARM/STM32/drivers/stm32-spi.ads
@@ -192,13 +192,20 @@ package STM32.SPI is
      (This     : in out SPI_Port;
       Incoming : out UInt8);
 
+   overriding
+   procedure Transfer
+     (This      : in out SPI_Port;
+      Data_Out  :        HAL.SPI.SPI_Data_8b;
+      Data_In   :    out HAL.SPI.SPI_Data_8b;
+      Status    :    out HAL.SPI.SPI_Status;
+      Timeout   :        Natural := 1000);
+
    procedure Transmit_Receive
      (This      : in out SPI_Port;
       Outgoing  : UInt8_Buffer;
       Incoming  : out UInt8_Buffer;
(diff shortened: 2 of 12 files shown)


Fabien-Chouteau:
Hello @evilspacepirate,

Let me check if I understand correctly. This means that a byte must be sent (even if it's zero) for every byte you want to receive. Is that only for the first byte or for all the bytes you may want to receive?

How is that done in other libraries like Arduino, Teezy or ARM CMSIS? 

Thanks,

evilspacepirate:
Fabien,

One of the properties of the SPI bus is that it is full duplex. During each
clock cycle the master sends a bit on the MOSI line and the slave reads it,
while the slave sends a bit on the MISO line and the master reads it.

To answer your question about SPI transfers in other libraries:

Arduino: transfer() https://www.arduino.cc/en/Reference/SPITransfer

Teezy: I couldn't find any information on this.

ARM CMSIS: This appears to support a full duplex 'Transfer' function:
https://github.com/ARM-software/CMSIS/blob/master/CMSIS/Driver/Include/Driver_SPI.h

Thanks,
John

Fabien-Chouteau:
Thanks John,

I forgot that SPI always send and receive at the same time :)

To be consistent with the rest of the HAL.SPI interface, the Transmit_Receive procedure should be replaced by something like this: 

```
procedure Transfer
     (This     : in out SPI_Port;
      Data_In  : SPI_Data_8b;
      Data_Out : out SPI_Data_8b;
      Status   : out SPI_Status;
      Timeout  : Natural := 1000) is abstract
     with
       Pre'Class => Data_Size (This) = Data_Size_8b;
```
The implementation will pad with zeros when Data_In and Data_Out have different lengths.
In STM32.SPI it should be possible to factorize Transmit and Receive to use the new Transfer but with empty Data_In or Data_Out arrays. I don't know if I'm clear :)

Regards,

evilspacepirate:
Fabien,

1. Why is a timeout parameter required for a SPI transfer operation?

2. For the Transfer procedure you specified, I think the implementation should transfer N number of bytes. Where N is the minimum length between the two parameters: Data_In and Data_Out (in case they have different lengths)

3. I would still like to request that we add a Transfer procedure that transfers exactly one byte of data. This is a primitive operation that does not require the overhead of run-time array length checks.

Thanks,
John

Fabien-Chouteau:
Hi John,

> 1. Why is a timeout parameter required for a SPI transfer operation?

Like for the other operations we have loops checking for flags, so the intent is to avoid being locked infinitely.

It's true that contrary to I2C, there is no multi-master mode in SPI. I don't know if we can assume that SPI transfer will always terminate in time...

>  1. For the Transfer procedure you specified, I think the implementation should transfer N number of bytes. Where N is the minimum length between the two parameters: Data_In and Data_Out (in case they have different lengths)

Why do you want to take the minimum length? It would mean that some data will not be transferred or received. Seems counter intuitive.

If you take the maximum length you can either pad with zero if receiving array is bigger or ignore the received data if sending array is bigger.

BTW, Data_In and Data_Out parameters are inverted in my proposal it should be:

```
procedure Transfer
     (This     : in out SPI_Port;
      Data_Out : SPI_Data_8b;
      Data_In  : out SPI_Data_8b;
      Status   : out SPI_Status;
      Timeout  : Natural := 1000) is abstract
     with
       Pre'Class => Data_Size (This) = Data_Size_8b;
```

>  1. I would still like to request that we add a Transfer procedure that transfers exactly one byte of data. This is a primitive operation that does not require the overhead of run-time array length checks.

Run-time checks can always be removed when performance is a concern, and given the complexity of those SPI sub-programs I don't think just the array checks will be significant.

I also want to avoid having too much primitives in the interface. We already have the 16 and 8 bit variants which seems redundant.

Regards,

evilspacepirate:
> Why do you want to take the minimum length? It would mean that some data will not be transferred or received. Seems counter intuitive.
> If you take the maximum length you can either pad with zero if receiving array is bigger or ignore the received data if sending array is bigger.

Extra documentation is required to describe this special behavior where the Data_Out and Data_In lengths are of different sizes. I think the simplest and most intuitive solution would be to force the Data_Out and Data_In arrays to have the same size using a precondition.

So something like:

    procedure Transfer
      (This     : in out SPI_Port;
       Data_Out : SPI_Data_8b;
       Data_In  : out SPI_Data_8b;
       Status   : out SPI_Status;
       Timeout  : Natural := 1000) is abstract
     with
       Pre'Class => Data_Size (This) = Data_Size_8b and
                    Data_Out'Length = Data_In'Length;


What do you think?

Fabien-Chouteau:
This could work but it's not as user friendly.

I agree that my proposal needs to be documented but it also feels pretty natural way to handle the case of different array lengths. 

I'll let you decide, the constraint can always be removed later if it makes sense.

evilspacepirate:
If the Ravenscar_Time package is not available, then how do we implement delays?

pat-rogers:
On 3/26/2017 3:50 PM, El Capitán wrote:
> If the Ravenscar_Time package is not available, then how do we implement delays?


We have a package coming that does it using the clock cycle counter.

Fabien-Chouteau:
Hi John,

> If the Ravenscar_Time package is not available, then how do we implement delays?

The problem here, is that a package of the component library is depending on the Ravenscar_Delay package. Here it's the body of your BlueNRG_MS package.

In your BlueNRG_MS_Device type you should have a `not null HAL.Time.Any_Delays`. See an example [here](https://github.com/AdaCore/Ada_Drivers_Library/blob/master/components/src/printer/adafruit_thermal_printer/adafruit-thermal_printer.ads). You can then use this Delay object in the body of the package.

Now that you are making progress I will try to do a more detailed review as soon as possible.

Regards,

PR #182, opened 2017-04-19
Remove submodule embedded-runtimes

lambourg:
This pull request prepares the coming changes in embedded runtimes installation, where runtimes will need to be installed before being able to use them.

To help transitioning the project files to the new schema, an automated project generator has been implemented to generate automatically the board projects. Those have been updated accordingly.

Finally, in order to test the changes, I tried to rebuild the examples and the STM32 driver demos, and came through various errors unrelated to the current change, but I still fixed those in the same branch, to ensure that this fix does not bring regressions.


lambourg:
For the record: starting with commit 1fa76dc, I managed to get rid of the aggregate projects usage in the board projects.

Pro: much nicer project hierarchy, make it easier to have board definitions outside of the Ada_Driver_Library repository, allow aggregate projects to use the ada drivers library board definitions.

Cons: we can't rely anymore on all object files to be in separate directories, depending on the choice of boards/rts/build mode. This leads to significant rebuilds when changing build mode or runtime profiles.

lambourg:
Note that I took the opportunity to rebase my demos on this branch and this lead to many regressions in the bitmap buffers and more generally the framebuffer support on the STM32 boards.

This should now be fixed, mostly.

In particular:
* the buffer x/y swapping was broken
* the framebuffer copy_back logic (when wanting to preserve the content of the hidden buffer within the fron buffer) was reverted

Also the performances of the graphics primitive are now improved, with careful allocation between dispatching calls vs. direct field access.

boards/gen.py
```diff
@@ -0,0 +1,216 @@
...
+    cnt += '\n'
+    cnt += '   for Source_Dirs use\n'
+    cnt += '     ("%s");\n' % '",\n      "'.join(board['source_dirs'])
+    cnt += '\n'
+    cnt += '   package Builder is\n'
+    cnt += '      for Switches ("Ada") use\n'
+    if board_name == 'Native':
+        cnt += '        ("-s");\n'
+    else:
+        cnt += '        ("--RTS=" & Project\'Runtime("Ada"), "-s");\n'
```
Fabien-Chouteau: Do we still need this when there's a Runtime attribute?

lambourg: > I don't think we should remove the submodule it's a nice way to get the dependency.

Well, we just can't keep this model. You know that.

lambourg:
>  I think there's a risk that we will not compile for each variant during testing, we will need one example for each.

We can try building the various board projects.

lambourg:
Closing, superseeded by #204

PR #183, opened 2017-04-25
[RFC] STM32 dma services

Fabien-Chouteau:
In this PR, I try provide a common scheme for DMA support in the STM32 communication drivers. 

First, STM32.DMA.Interrupts factorizes the handling of interrupts for DMA transfer. It's a simple protected object, the configuration of DMA streams is the same.

I then use this in the STM32.SPI.DMA driver in a proof of concept that can be applied to I2C, UART, I2S, etc. The idea is that the configuration code is shared, only the transfers primitives differ which makes for a simple and clear implementation.

These new packages are used in OpenMV.Sensor for the DCMI transfer and OpenMV.LCD with SPI port. 

Comments are welcome :)

### DIFF
--- a/arch/ARM/STM32/devices/stm32f42x/stm32-device.adb
+++ b/arch/ARM/STM32/devices/stm32f42x/stm32-device.adb
@@ -436,7 +436,7 @@ package body STM32.Device is
    -- Enable_Clock --
    ------------------
 
-   procedure Enable_Clock (This : SPI_Port) is
+   procedure Enable_Clock (This : SPI_Port'Class) is
    begin
       if This.Periph.all'Address = SPI1_Base then
          RCC_Periph.APB2ENR.SPI1EN := True;
@@ -459,7 +459,7 @@ package body STM32.Device is
    -- Reset --
    -----------
 
-   procedure Reset (This : SPI_Port) is
+   procedure Reset (This : SPI_Port'Class) is
    begin
       if This.Periph.all'Address = SPI1_Base then
          RCC_Periph.APB2RSTR.SPI1RST := True;
--- a/arch/ARM/STM32/devices/stm32f42x/stm32-device.ads
+++ b/arch/ARM/STM32/devices/stm32f42x/stm32-device.ads
@@ -44,17 +44,21 @@
 
 with STM32_SVD;     use STM32_SVD;
 
-with STM32.DMA;     use STM32.DMA;
-with STM32.GPIO;    use STM32.GPIO;
-with STM32.ADC;     use STM32.ADC;
-with STM32.USARTs;  use STM32.USARTs;
-with STM32.SPI;     use STM32.SPI;
-with STM32.I2S;     use STM32.I2S;
-with STM32.I2C;     use STM32.I2C;
-with STM32.Timers;  use STM32.Timers;
-with STM32.DAC;     use STM32.DAC;
-with STM32.RTC;     use STM32.RTC;
-with STM32.CRC;     use STM32.CRC;
+with Ada.Interrupts.Names;
+
+with STM32.DMA;            use STM32.DMA;
+with STM32.DMA.Interrupts; use STM32.DMA.Interrupts;
+with STM32.GPIO;           use STM32.GPIO;
+with STM32.ADC;            use STM32.ADC;
+with STM32.USARTs;         use STM32.USARTs;
+with STM32.SPI;            use STM32.SPI;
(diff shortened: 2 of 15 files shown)


Fabien-Chouteau:
> Shouldn't the designated type be class-wide? That is expected with the name Any_*

You mean for Any_DMA_Interrupt_Controller? It's not possible because it's not a tagged or interface type.

pat-rogers:
On 5/9/2017 9:44 AM, Fabien Chouteau wrote:
>> Shouldn't the designated type be class-wide? That is expected with the name Any_*
>
> You mean for Any_DMA_Interrupt_Controller? It's not possible because it's not a tagged or interface type.

In that case we should use a different name, since Any_* means 
class-wide, no?

Fabien-Chouteau:
> In that case we should use a different name, since Any_* means class-wide, no?

Since class-wide doesn't apply here, I think Any_* is good for the access all types whether or not they are class-wide.

Any_DMA_Interrupt_Controller is indeed an access to any DMA_Interrupt_Controller because there won't be any derived type.

pat-rogers:
On 5/9/2017 11:08 AM, Fabien Chouteau wrote:
>> In that case we should use a different name, since Any_* means class-wide, no?
>
> Since class-wide doesn't apply here, I think Any_* is good for the access all types whether or not they are class-wide.
>
> Any_DMA_Interrupt_Controller is indeed an access to any DMA_Interrupt_Controller because there won't be any derived type.


Seems potentially confusing without a compelling justification... mumble...

pat-rogers:
On 5/9/2017 11:08 AM, Fabien Chouteau wrote:
>> In that case we should use a different name, since Any_* means
>> class-wide, no?
>
> Since class-wide doesn't apply here, I think Any_* is good for the
> access all types whether or not they are class-wide.
>
> Any_DMA_Interrupt_Controller is indeed an access to any
> DMA_Interrupt_Controller because there won't be any derived type.

Since a class-wide type represents a set of types, ie, potentially 
multiple members, that's a different situation.

Our convention when designating such a set is to name the access type 
Any_* because the access value can designate any of the several members 
of that set.

Using that same convention for a case that will never represent a set is 
potentially misleading to the reader and IMO would require a compelling 
reason. The value of a convention is consistency in what it tells the 
reader, so this seems like an unjustified deviation.

(Yes, we're back to the subject of naming again, hence my soapbox. :-)

Fabien-Chouteau:
> Using that same convention for a case that will never represent a set is potentially misleading to the reader and IMO would require a compelling reason. The value of a convention is consistency in what it tells the reader, so this seems like an unjustified deviation.

I can agree with that, but it means we need another naming convention for access to non class-wide types.

pat-rogers:
On 5/15/2017 5:06 AM, Fabien Chouteau wrote:
>> Using that same convention for a case that will never represent a set is potentially misleading to the reader and IMO would require a compelling reason. The value of a convention is consistency in what it tells the reader, so this seems like an unjustified deviation.
> 
> I can agree with that, but it means we need another naming convention for access to non class-wide types.


How about *_Reference, when the type is visibly an access type (ie not a 
private type)?

Fabien-Chouteau:
> How about *_Reference, when the type is visibly an access type (ie not a private type)?

It's a bit long, maybe _Ref or _Access or _Acc or _Ptr...

pat-rogers:
On 5/18/2017 8:33 AM, Fabien Chouteau wrote:
>> How about *_Reference, when the type is visibly an access type (ie not a private type)?
> 
> It's a bit long, 
Programs are read far more times than they are written, so the software 
engineering guideline is to favor the reader over the convenience of the 
writer.

Fabien-Chouteau:
> Programs are read far more times than they are written, so the software engineering guideline is to favor the reader over the convenience of the writer.

I agree with that, and it seems that we disagree on readability of _Ref vs _Reference.

If you don't want shorten words then I think _Access is more appropriate than _Reference.

pat-rogers:
On 5/18/2017 11:02 AM, Fabien Chouteau wrote:
>> Programs are read far more times than they are written, so the software engineering guideline is to favor the reader over the convenience of the writer.
> 
> I agree with that, and it seems that we disagree on readability of _Ref vs _Reference.

When reading it, in my head I pronounce "Ref" as "Reference" so I think 
it might as well be spelled out.

> 
> If you don't want shorten words then I think _Access is more appropriate than _Reference.

I can live with that. Not much difference.

Fabien-Chouteau:
OK, that's the middle ground then.

Since there's no comment on the substance I will continue this way.

PR #184, opened 2017-05-02
Various fixes

lambourg:
This fixes various issues I discovered while moving the demos to master.

* STM32.Setup: automatically calling this function fails if two I2C peripherals are used. This is the case for example when using the Touch Panel and the Audio drivers. This pull request checks that the port is uninitialised before trying to re-initialising it.
* HAL.Audio: fix the alignment of the Audio buffer. By default, it should be at least 16-bit aligned (for compatibility with DMA transfers).

### DIFF
--- a/arch/ARM/STM32/drivers/stm32-setup.adb
+++ b/arch/ARM/STM32/drivers/stm32-setup.adb
@@ -45,6 +45,10 @@ package body STM32.Setup is
    is
       I2C_Conf : I2C_Configuration;
    begin
+      if STM32.I2C.Port_Enabled (Port) then
+         return;
+      end if;
+
       --  GPIO --
       Enable_Clock (SDA & SCL);
 
--- a/hal/src/hal-audio.ads
+++ b/hal/src/hal-audio.ads
@@ -34,7 +34,7 @@ with Interfaces; use Interfaces;
 package HAL.Audio is
 
    type Audio_Buffer is array (Natural range <>) of Integer_16
-     with Component_Size => 16, Alignment => 16;
+     with Component_Size => 16, Alignment => 2;
 
    type Audio_Volume is new Natural range 0 .. 100;
 


Fabien-Chouteau:
> STM32.Setup: automatically calling this function fails if two I2C peripherals are used. This is the case for example when using the Touch Panel and the Audio drivers. This pull request checks that the port is uninitialised before trying to re-initialising it.

Calling the initialization twice is most likely an error, in that case an exception would be more helpful than a silent return.

lambourg:
Ok, so what do you propose? Currently, using the Audio and the Touch Screen raise an exception because the I2C is initialised twice. That was not the case before the introduction of STM32.Setup.

Fabien-Chouteau:
> Ok, so what do you propose? Currently, using the Audio and the Touch Screen raise an exception because the I2C is initialised twice. That was not the case before the introduction of STM32.Setup.

Make sure at the board support level that setup is called only once. This will also have the benefit or removing a couple of duplicated lines of code.

lambourg:
Well, that's exactly what this change is about. So what do you propose further (e.g. the STM32.Setup introduction is on your behalf, so what do you see as a fix to the current issue?)

Fabien-Chouteau:
> Well, that's exactly what this change is about. So what do you propose further

No, your change is to silently return when trying to reconfigure a I2C port with potentially incompatible GPIOs or speed. Creating a pitfall very hard to debug.

> (e.g. the STM32.Setup introduction is on your behalf, so what do you see as a fix to the current issue?)

I said "at the board level". Add an init procedure in STM32.Board than can be called by both audio and TP but will do the init only once. Not only this will factorize the code, it will make it more explicit that this I2C port is a shared resource and should be treated as such.

Fabien-Chouteau (approved):

PR #185, opened 2017-05-08
restore GPIO handling performance

pat-rogers:
Go back to bit-oriented rather than array oriented accesses to take advantage of hardware designed for that purpose.


pat-rogers:
Requires later version of compiler.

PR #187, opened 2017-05-08
restore gpio bit set/reset performance; reinstate basic gpio demo

pat-rogers:
The bit-reset registers ignore writes of zeros. thus we can can use bit masks and not worry about
preserving exiting set/cleared bits. This leads to the maximally efficient code generated. 

remove inlining for routines that the compiler will not inline (in later versions of the compiler).

reinstate basic gpio demo.


arch/ARM/STM32/drivers/stm32-gpio.adb
```diff
@@ -78,8 +78,9 @@ package body STM32.GPIO is
 
    overriding
    function Mode (This : GPIO_Point) return HAL.GPIO.GPIO_Mode is
+      Index : constant Integer := GPIO_Pin'Pos (This.Pin);
```
Fabien-Chouteau: Not blocking but the index type should probably be GPIO_Pin_Index.

arch/ARM/STM32/driver_demos/demo_gpio_direct_leds/demo_gpio.gpr
```diff
@@ -0,0 +1,53 @@
+with "../../../../../boards/common_config.gpr";
```
Fabien-Chouteau: This test should be added `the scripts/build_all_examples.py` to make sure we don't break it in the future.

pat-rogers:
Will resubmit with changes

PR #190, opened 2017-05-13
Add dwt for m0 and Ada interface for all

pat-rogers:


arch/svd.mk
```diff
@@ -1,4 +1,3 @@
-# IMPORTANT NOTE: due to a bug in the compiler (Q207-001), the data structure in stm32_svd-crc.ads must be manually modified after this makefile
 # generates the code. See Ada Drivers Library issue #163.
```
Fabien-Chouteau: This line should probably be removed as well

PR #191, opened 2017-05-14
nRF51: Add timer support

Fabien-Chouteau:


arch/ARM/Nordic/drivers/nrf51-timers.ads
```diff
@@ -0,0 +1,158 @@
...
+   -- Configuration --
+   -------------------
+
+   procedure Set_Prescaler (This      : in out Timer;
+                            Prescaler : UInt4);
+   --  Set timer prescaler value
+   --
+   --  Timer frequency = HFCLK / (2**Prescaler)
+   --
+   --  This procedure should only be called when the timer is stopped.
```
pat-rogers: Would be a good precondition.

Fabien-Chouteau: Yes, I tried to do it but I don't think there's any reliable way to know if the timer is stopped or not...

PR #194, opened 2017-05-19
Minor fixes in the BNO055 IMU component.

lambourg:


components/src/motion/bno055/bosch_bno055.adb
```diff
@@ -239,14 +239,13 @@ package body Bosch_BNO055 is
      (This : in out BNO055_9DOF_IMU;
       Mode : Operating_Modes)
    is
-      --  These delay values do not match the Datasheet, table 3-6, but they
-      --  are what the Bosch driver uses
-      Switching_From_Config_Mode : constant := 600; -- milliseconds
+      --  Table 3-6 of the datasheet + 1ms as margin
+      Switching_From_Config_Mode : constant := 8; -- milliseconds
       Switching_To_Config_Mode   : constant := 20; -- milliseconds
```
pat-rogers: Is there a reason to believe the Datasheet value, over that of the Bosch driver code?

components/src/motion/bno055/bosch_bno055.adb
```diff
@@ -239,14 +239,13 @@ package body Bosch_BNO055 is
...
    begin
       if This.Mode = Operating_Mode_Config then
          Write (This.Port, BNO055_OPR_MODE_ADDR, Value => Mode'Enum_Rep);
-         Delay_Milliseconds (Switching_From_Config_Mode);
+         Delay_Milliseconds (Switching_To_Config_Mode);
```
pat-rogers: That's wrong though. The if-statement expression, when true, says we are already in the config mode, so the change is from that mode, rather than into it.

lambourg:
> Is there a reason to believe the Datasheet value, over that of the Bosch driver code?

Well, it works from what I can see, and 600ms made the driver very slow at init.

Maybe a typo or a leftover in the Bosch driver?

pat-rogers:
On 5/19/2017 9:52 AM, Lambourg Jérôme wrote:
>> Is there a reason to believe the Datasheet value, over that of the Bosch driver code?
> 
> Well, it works from what I can see, and 600ms made the driver very slow at init.
> 
> Maybe a typo or a leftover in the Bosch driver?

Maybe so.  OK, let's go with it. I'll try the lunar lander demo with the 
change too.

PR #211, opened 2017-07-28
Implement logging tools with priority and categories handling

Fabien-Chouteau:


Fabien-Chouteau:
> Maybe function Pop could be Popped or something similar...

Or it could just be a procedure because the argument out Length will already tell if there's nothing in queue.

pat-rogers:
Pat
> On Aug 7, 2017, at 4:23 AM, Fabien Chouteau <notifications@github.com> wrote:
> 
> Maybe function Pop could be Popped or something similar...
> 
> Or it could just be a procedure because the argument out Length will already tell if there's nothing in queue.
> 

That would be better indeed.

PR #239, opened 2017-12-20
add convenience routine to conditionally set/clear a pin

pat-rogers:


pat-rogers:
I was going to name the procedure Set, like the others, but that didn't seem right because the pin is cleared if the condition is False...

PR #247, opened 2018-03-01
adding sudoku solver using Display on stm32f429_Disco

stangassinger:


stangassinger:
The sudoku solver is a example where the stm32f429_Disco is used with a simple GUI to solve a sudoku.

Fabien-Chouteau:
Hi @stangassinger,

Thanks for you contribution!

I think we already have enough examples for this board so it's probably better if you to create a separate repository for your sudoku project. And then we can add it to the list of project that we have in the README like we did for the other projects that use the Ada Drivers Library.

Regards,

PR #248, opened 2018-03-03
Update blinky_f4disco.gpr

plestedr:
Change object dir for 407 Disco to point to the 407 tree, not the 429's object dir.


Fabien-Chouteau:
Thanks @plestedr !

PR #250, opened 2018-03-07
Bring back the IO Functions definitions.

elbric0:
I suggest to bring back the IO Functions definitions. I think they are useful because they avoid us to open the MCU doc to check what is the IOF number for a particular peripheral and I think they are more meaningful when reading the code. Example: If I need to assign a pin to the UART, I prefer to write Set_IO_Function (P16, IOF_UART0); than Set_IO_Function (P16, IOF0);.


Fabien-Chouteau:
Thanks @elbric0 !

PR #265, opened 2018-07-11
[RFC] HAL.GPIO: Switch to a capability system

Fabien-Chouteau:
Instead of returning False when a configuration is not supported, users
can call a function beforehand to know if it is supported. The
capability is also now a precondition of the setup procedures.


Fabien-Chouteau:
> (And I like the functions changed to procedures, of course :)

Of course :)

Thanks Pat

Issue #267, opened 2018-07-12
Microbit Button_B not working


simonjwright:
In `boards/MicroBit/src/microbit-buttons.adb`, function `State`, we have
```
      if Button =  Button_B then
         --  For a reason that I don't understand right now, the B button is a
         --  always detected as pressed (low). As far as I can see the code is
         --  the same as button A and probing the hardware didn't show any
         --  problem...
         raise Program_Error with "Button B is not working...";
      end if;
```
I believe I’ve found the reason: in `microbit-time.adb`, procedure `Initialize`, we have
```
         Clocks.Set_Low_Freq_Source (Clocks.LFCLK_XTAL);
```
(weird, there isn’t a crystal attached to the low-frequency inputs at all; the only crystals on the board are 16 MHz, not the 32 kHz you’d need for the low frequency source), and this has the effect of what on STM chips would be called an alternate function on that pin, so that GPIO pin 26 is effectively disconnected.

Fix:
```
         Clocks.Set_Low_Freq_Source (Clocks.LFCLK_SYNTH);
```
(the HF clock appears to run without needing any configuration).

PR #268, opened 2018-07-13
MicroBit: Fix button B

Fabien-Chouteau:
As discussed in #267, MicroBit.Time was configuring the Low Frequency
Clock to use an external crystal, but there is no LFC external crystal
on the micro:bit. Also the button B is connected to the pin used for the
LFC crystal, so this configuration meant the pin was actually not
available as GPIO for the button code.

MicroBit.Time now uses synthesized clock for LFC so the pin is available
for button B.

### DIFF
--- a/boards/MicroBit/src/microbit-buttons.adb
+++ b/boards/MicroBit/src/microbit-buttons.adb
@@ -97,13 +97,6 @@ package body MicroBit.Buttons is
 
    function State (Button : Button_Id) return Button_State is
    begin
-      if Button =  Button_B then
-         --  For a reason that I don't understand right now, the B button is a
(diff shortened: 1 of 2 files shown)


Fabien-Chouteau:
Ping @simonjwright

simonjwright:
Great, that was quick! (I assume you have a test program? I don’t, not in Ada_Drivers_Library, I was working on [cortex-gnat-rts](https://github.com/simonjwright/cortex-gnat-rts/blob/master/microbit/adainclude/nrf51_clock.adb#L50))

Fabien-Chouteau:
I'm actually working on micro:bit examples, so I'll add one for the buttons.

PR #270, opened 2018-07-19
Added Ide info to micro:bit text/button GPR files

mosteo:
Without those the "Flash to board" button does not appear in GPS toolbar. 
Copied from the example project that comes with GPS 2018.


Fabien-Chouteau:
Thanks @mosteo, do you mind doing this for all the examples?

Issue #271, opened 2018-07-25
Style check only on pull requests


Fabien-Chouteau:
Björn made a good comment on [comp.lang.ada](https://groups.google.com/forum/#!topic/comp.lang.ada/OidrH33UBJg) that enforcing the coding style with error message (`-gnaty -gnatwe`) is not friendly for beginners starting to use the library.

In the end we only want to enforce the coding style on code that is checked-in. So there might be a way to only enable style check for pull requests validation.

yannickmoy:
why? This only applies to the library itself, not the user code.

Fabien-Chouteau:
It also applies to the examples that users will start from.

PR #273, opened 2018-09-03
Optional style checks and warning as error switches

Fabien-Chouteau:
Enforcing style checks only in Production mode doesn't lower the quality
of the code base since all contributions have to compile in Production
before they can be merged (enforced by CI and GitHub rules).

On the other hand disabling style checks for debug mode means it is
easier and more friendly to play with the library and the examples.

Part of #271


pmderodat:
This feels a bit backwards: it’s considered good practice not to abort production builds because of style checks or warnings, as it may break the compilation of one release in the future with a new compiler… So unless the CI builds in a special mode (Prod + error on warnings/stylechecks), you lose these checks.

Isn’t it possible instead to somehow tune project files so that examples are not built in production mode but the library itself is?

Fabien-Chouteau:
New version:

This patch makes style checks and warning as error switches controllable
with a scenario variable: ADL_BUILD_CHECKS (disabled by default).

With this we can enforce style checks on all contributions by enabling
ADL_BUILD_CHECKS in testsuite runs and CI builds.

On the other hand the checks are disabled by default which means it is
easier and more friendly to play with the library and the examples.

Fabien-Chouteau:
The AppVeyor build timed out once again...

I think we have to disable the dual builds of Debug and Production modes.

PR #276, opened 2018-09-18
Don't use DMA Half Transfer Complete

simonjwright:
The previous attempt at this solution (#274) gave the user too much control over what was intended to be an implementation for the general user. This version eliminates the observed problem (the `Half_Transfer_Complete_Indicated` status wasn’t always cleared, resulting in a postcondition failure) by not enabling the Half Transfer Complete interrupt and not checking the corresponding status (note, the status will become set regardless of whether the interrupt is enabled).

  * arch/ARM/STM32/drivers/dma_interrupts/stm32-dma-interrupts.adb
    (Start_Transfer): don't enable Half_Transfer_Complete_Interrupt.
    (Interrupt_Handler): don't check/clear Half_Transfer_Complete_Indicated.

### DIFF
--- a/arch/ARM/STM32/drivers/dma_interrupts/stm32-dma-interrupts.adb
+++ b/arch/ARM/STM32/drivers/dma_interrupts/stm32-dma-interrupts.adb
@@ -1,6 +1,6 @@
 ------------------------------------------------------------------------------
 --                                                                          --
---                    Copyright (C) 2016-2017, AdaCore                      --
+--                    Copyright (C) 2016-2018, AdaCore                      --
 --                                                                          --
 --  Redistribution and use in source and binary forms, with or without      --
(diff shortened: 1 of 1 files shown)


Fabien-Chouteau:
Thanks @simonjwright this looks like a good solution. Did you test it on the Crazyflie?

simonjwright:
@Fabien-Chouteau said
>Did you test it on the Crazyflie?

Yes, indeed! Without the "ignore the status", and with `ADA_BUILD` set to `Debug`, the postcondition failed within a minute or two (the transfer count was 6). With "ignore the status", the Crazyflie ran for a couple of hours.

pat-rogers:
On 9/18/2018 11:25 AM, Fabien Chouteau wrote:
> Thanks a lot @simonjwright <https://github.com/simonjwright> !

Yes indeed!

PR #277, opened 2018-09-26
Add minimal example of servo motor control.

quinot:


boards/MicroBit/src/microbit-servos.ads
```diff
@@ -0,0 +1,52 @@
...
+------------------------------------------------------------------------------
+
+with MicroBit.IOs; use MicroBit.IOs;
+
+package MicroBit.Servos is
+
+   --  Servo-motors control
+
+   subtype Servo_Pin_Id is Pin_Id
+     with Predicate => Supports (Servo_Pin_Id, Analog);
```
Fabien-Chouteau: That's a cool use of the predicate :+1:

PR #278, opened 2018-10-05
Ensure value of State is correct even in state change callbacks

quinot:
Update the States array prior to calling button state change callbacks,
so that it reflects the current, not previous, state of buttons.


Fabien-Chouteau:
Thanks @quinot

PR #298, opened 2019-03-27
[WIP] E31 Arty FPGA Evaluation Kit

borancar:
Please do not merge, this is only meant to start a discussion. I've been trying to port the example to SiFive's E31 FPGA Evaluation Kit, available at https://dev.sifive.com after a free registration. This is the first stop for people wishing to evaluate SiFive on FPGA boards. So far I've succeeded in porting blinky across.

I rebased the work extensively into distinct commits that try to address some of the obstacles I've come across. Ideally, only the last 2 commits should be part of a PR that addresses adding a new board for RISC-V. I would like to help here, if you agree.

Obstacles I've had trouble with:
- In svd2ada, I needed to duplicate FE310.svd which results in a whole set of extra files that are unnecessary as SiFive devices are the same, no matter what board. The commit "E31 svd2ada redundant" (https://github.com/AdaCore/Ada_Drivers_Library/pull/298/commits/d2ab605886a28f3b2955af71483e224ab9e1ba22) tries to show the difference by overwriting files copied with rename from FE310. Note that some of the non-trivial differences were a result of svd2ada updates done since the files were last generated (PLIC/CLINT/UART0_Base instead of hardcoded addresses). I couldn't find an easy way of separating the peripherals from the actual board. Ideally, only the commit "Run svd2ada for E31 Arty FPGA Kit" (https://github.com/AdaCore/Ada_Drivers_Library/pull/298/commits/b9a136ec09ed050384a439071f8d290ac18d4c5d) should be needed to add SVD support for a RISC-V board based on a standard core.
- In spirit of the previous one, I needed to copy all the drivers with a rename for the E31, as shown in the commit "Modify drivers for E31" (https://github.com/AdaCore/Ada_Drivers_Library/pull/298/commits/dbf4e4dd269d2ddd9c2776bfe188c2b49c9db9ec). All the changes are trivial here. Is there a better way of extracting these drivers as generic SiFive drivers that only need a base address and an IRQ number passed to them, in the spirit of device-tree files standard in the Linux kernel and also something that SiFive uses for their embedded targets with Metal?
- I needed to add a new runtime, zfp-e31arty. I've achieved this just by copying and trivially modifying zfp-hifive1 (file locations and flash addresses) in the GNAT Community Edition. I didn't care about TextIO at this early stage, but was wondering if it's possible to modify https://github.com/AdaCore/bb-runtimes/blob/community-2018/riscv/sifive/hifive1/s-textio.adb so that it relies on something defining the UART later on, similar to how newlib allows you to define _read, _write and other IO functions?


Fabien-Chouteau:
Hi @borancar,

> This is the first stop for people wishing to evaluate SiFive on FPGA boards. So far I've succeeded in porting blinky across.

That's great!

> I rebased the work extensively into distinct commits that try to address some of the obstacles I've come across. Ideally, only the last 2 commits should be part of a PR that addresses adding a new board for RISC-V. I would like to help here, if you agree.

Thanks for your help. Based on your comments, we need to find a good solution to have more flexibility in the drivers. 

> Obstacles I've had trouble with:
> 
>     - In svd2ada, I needed to duplicate FE310.svd which results in a whole set of extra files that are unnecessary as SiFive devices are the same, no matter what board.

This is the most annoying limitation of the SVD format in my opinion, it is not modular.
I hope that the companies like SiFive will come up with an improved version to tackle this problem.
 
Maybe we can go away from the the monolithic SVD file and make an SVD for each peripheral?

> Is there a better way of extracting these drivers as generic SiFive drivers that only need a base address and an IRQ number passed to them, in the spirit of device-tree files standard in the Linux kernel and also something that SiFive uses for their embedded targets with Metal?

Ada generics are quite powerful so it is worth trying.

>     - I needed to add a new runtime, zfp-e31arty. I've achieved this just by copying and trivially modifying zfp-hifive1 (file locations and flash addresses) in the GNAT Community Edition. 

We are working on a solution for this. The plan is for the users to be able to generate the linker script and the crt0 so that the run-time will be the same on any RISC-V32.

> I didn't care about TextIO at this early stage, but was wondering if it's possible to modify https://github.com/AdaCore/bb-runtimes/blob/community-2018/riscv/sifive/hifive1/s-textio.adb so that it relies on something defining the UART later on, similar to how newlib allows you to define _read, _write and other IO functions?

On ARM cortex-M we always use semi-hosting, because it is standard across all the implementations.Do you know the status of semi-hosting definition for RISC-V?

Regards,

PR #304, opened 2019-05-27
L6470h

avraliov:
Stm l6470h stepper motor driver

### DIFF
--- a/components/src/motion/l6470h/l6470h.adb
+++ b/components/src/motion/l6470h/l6470h.adb
@@ -38,8 +38,6 @@ package body L6470h is
    begin
       Next_Release := Next_Release + Period;
       delay until Next_Release;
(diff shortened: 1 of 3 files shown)


Fabien-Chouteau:
Hello @avraliov , thanks for your contribution.

The code as is cannot be included in Ada Drivers Library because your component `l6470h` depends on MCU specific packages (`STM32.Timers`, `STM32.Device`, `STM32.SPI`, etc/). As described in this document: [design.md](https://github.com/AdaCore/Ada_Drivers_Library/blob/master/docs/design.md) the components must be platform agnostic and therefore only depend on `HAL` packages.

Here is an example of how this can be done: [st7735r.ads](https://github.com/AdaCore/Ada_Drivers_Library/blob/master/components/src/screen/ST7735R/st7735r.ads)

PR #308, opened 2019-08-20
Add support for startup-gen

Fabien-Chouteau:
startup-gen is a tool that generates crt0 and linker script from a
device configuration (CPU name, memory layout). The tool only works for
ZFP run-times right now.

This patch also switches the HiFive1 support to use startup-gen.


Fabien-Chouteau:
Thanks @pat-rogers

PR #311, opened 2019-09-16
 MicroBit: Add console package

Fabien-Chouteau:
Provides an interface to send and receive character on the UART over USB
line of the MicroBit.


Fabien-Chouteau:
Thanks @quinot :)

PR #314, opened 2019-11-03
micro:bit: add an example for the accelerometer

Fabien-Chouteau:
The accelerometer init procedure is replaced by elaboration initialization to
match the interface of the other micro:bit features.


crisdeodates:
Fabien, Thank you for adding this example.

PR #315, opened 2019-11-14
Switch continuous integration to GitHub Action workflow

Fabien-Chouteau:


Fabien-Chouteau:
Looks like GitHub actions are not available for us yet.

Issue #318, opened 2019-12-07
scripts/install_dependencies.py is unable to checkout the community-2019 branch


WickedShell:
Following the `docs/how_to_create_my_own_adl_project.md` on making a standalone project. When installing the runtimes it will fail as the branch already exists:

```
wickedshell@milton ~/code % mkdir project_root_dir
wickedshell@milton ~/code % cd project_root_dir 
wickedshell@milton ~/code/project_root_dir % git clone https://github.com/AdaCore/Ada_Drivers_Library.git Ada_Drivers_Library/repo/
Cloning into 'Ada_Drivers_Library/repo'...
remote: Enumerating objects: 152, done.
remote: Counting objects: 100% (152/152), done.
remote: Compressing objects: 100% (97/97), done.
remote: Total 20303 (delta 55), reused 120 (delta 50), pack-reused 20151
Receiving objects: 100% (20303/20303), 18.06 MiB | 15.21 MiB/s, done.
Resolving deltas: 100% (13376/13376), done.
wickedshell@milton ~/code/project_root_dir % ./Ada_Drivers_Library/repo/scripts/install_dependencies.py
ROOT_DIR :/home/wickedshell/code/project_root_dir/Ada_Drivers_Library/repo
$ git clone https://github.com/AdaCore/bb-runtimes /home/wickedshell/code/project_root_dir/Ada_Drivers_Library/repo/bb-runtimes

$ git -C /home/wickedshell/code/project_root_dir/Ada_Drivers_Library/repo/bb-runtimes checkout -b community-2019 origin/community-2019

git branch checkout error (returned 128):
fatal: A branch named 'community-2019' already exists.

Running build command:
$ python /home/wickedshell/code/project_root_dir/Ada_Drivers_Library/repo/bb-runtimes/install.py --arch=arm-eabi

Dependency install command error (returned 1):
  File "/home/wickedshell/code/project_root_dir/Ada_Drivers_Library/repo/bb-runtimes/install.py", line 20
    print "usage: install.py [--arch=arm-eabi|aarch64-elf] [--prefix=<path>]"
                                                                            ^
SyntaxError: Missing parentheses in call to 'print'. Did you mean print("usage: install.py [--arch=arm-eabi|aarch64-elf] [--prefix=<path>]")?
```

[...]

PR #319, opened 2019-12-08
Pass the python environment down

WickedShell:
This solves switching to Python 3 by accident when installing the dependencies, as mentioned in #318. This doesn't fix checking out the already existing branch though.

The newline removals were just cleanups I cherry picked as when I turned on some of the GNAT style checks they were causing the example to fail.

I signed and sent the CLA in today.


Fabien-Chouteau:
Thank you @WickedShell !

PR #320, opened 2019-12-09
install_dependencies.py

Fabien-Chouteau:
Now that community-2019 is the default branch for the bb-runtimes repo,
we don't have to switch to it anymore.


WickedShell:
Fixed the problem for me, thanks :)

PR #327, opened 2020-02-04
Examples: Simplify lcd_std_out character emission

WickedShell:
This was an old cleanup I had from playing with some of the examples. `Internal_Put` was duplicating the logic that `New_Line` contains for advancing the line on the screen.


WickedShell:
I did sign the CLA, but I e-mailed it in per the documents on it rather then via CLA bot originally, I did it before sending #319 in, and would rather not add it into the CLA hook again if I don't have to.

Fabien-Chouteau:
Thank you @WickedShell :)

PR #328, opened 2020-02-11
Rename cm0 NVIC api functions to match the cm4_cm7 impl

nocko:
This allows for a unified driver in families that have members on both
sides. Useful for nRF family, but also should lower-end STM32 parts become
supported in the future.


Fabien-Chouteau:
Thank you @nocko !

PR #329, opened 2020-02-13
Changes cm0 Interrupt_Priority to UInt32 subtype for compat w/ cm4_7

nocko:
This would allow removal of last NVIC glue in the nRF family, but it does so with a little less safety.

Currently cm0 Interrupt_Priority is UInt8, in this patch it's a range subtype on UInt32. It would be a warning at compile time to specify an out-of-range value, but currently it's a compile time error.


nocko:
I'd really like something like a static assertion for this, but Preconditions seem to be runtime. Is there a better way to do this?

nocko:
Checked the cm4_7 code again. I'm going to rethink this.

PR #330, opened 2020-02-13
Fixes broken interrupt prio on cm0, now compat w/ cm4_7

nocko:
In my travels trying to make the cm4_7 and cm_0 NVIC APIs compatible, I found out that the cm0 Set_Priority was broken. It neglected to shift left the priority as required, so for the microbit examples, all interrupts were priority 0 (the least significant 6 bits of the prio ignore writes).

The Set_Priority API is compatible between implementations now, an as a bonus, the cm_0 implementation will warn if you specify an out of range priority.

### DIFF
--- a/arch/ARM/Nordic/drivers/nrf51-interrupts.adb
+++ b/arch/ARM/Nordic/drivers/nrf51-interrupts.adb
@@ -29,8 +29,8 @@
 --                                                                          --
 ------------------------------------------------------------------------------
 
-with Cortex_M.NVIC;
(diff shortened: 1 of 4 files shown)


nocko:
@Fabien-Chouteau You've addressed most of this in your work yesterday, I'll rebase the bugfix on master once those changes land.

Fabien-Chouteau:
Hello @nocko, can you rebase your change?

Fabien-Chouteau:
Fixed in #373

Issue #334, opened 2020-03-11
HiFive1 rev B is not fully compatible with the original HiFive1


Irvise:
**Summary:** While the HiFive1 rev B is just like its predecessor, the boot sequence has been changed and the tooling to upload the .hex files.

**Introduction:** I recently got my hands on the HiFive1 rev B and learned that there were Ada drivers for it. Upon trying them, they failed. A bit of research shows the original version has a bootloader that jumps to 0x20400000 to start executing code. The new version (rev B) changed the bootloader to jump to 0x20100000. The new model also has less internal memory (from 128Mbit down to 32Mbit), 2 UARTs and 1 I2C, a low-power state, optional bluetooth and wifi (which come by default if the boards is bought from Crowdsupply), uploading mechanism and an updated version of the debugging interface (version 0.13).

In order to make Ada work, even in a rough state, I changed the HiFive configuration (the link.ld and the zfp.gpr) to set the origin of the elf to 0x20100000. Which made the resulting binary work (after manually transforming the elf to hex). The upload method used was just copying the resulting .hex to the "USB" entry in the file manager, since the new rev B version only allows for JLink and copying into memory (OpenOCD is not officially supported, but there are people working on it, see https://forums.sifive.com/t/progress-getting-openocd-usable-with-revb/3415)

I have seen that the file /scripts/config/boards.py also contains the 0x204... address. I did not change it (I actually forgot) but the project still runs fine. Therefore, I don't know if the Python script is used to build the project or not. After all, it seems to contain generic information about the boards 

**Conclusion:** while the modification presented allows Ada to build working binaries, it is quite unsatisfactory. I believe there are enough differences between HiFive1 and HiFive1 rev B that a new board entry is needed, since selecting the boot entry with a build flag could potentially "break" a board. In this new entry, the boot address, upload mechanism and other hardware modifications could be added. Obviously, the new entry would be based on the original. **Notice:** this is just what I found in a day of fiddling with the board, there may be more issues.

However, my knowledge of embedded programming only extends to Arduino boards and PICs, which may not be ideal for me to handle this correctly. Furthermore, I am just getting started with Ada, so there is not much I can do for now (well, I could try different things, but I doubt they will have the necessary quality standards).

[...]

PR #335, opened 2020-03-11
Add support for the HiFive1 rev B

Fabien-Chouteau:
@Irvise can you try this and tell me if it works.

Related to #334.


Irvise:
I have built and uploaded the example project as described in #334.

It did not have any effect. As in, the program that I had uploaded earlier was not overwritten.

The program that I had was the "hello world"  that Freedom Studio includes. In this [gist](https://gist.github.com/Irvise/d8b3bf428a9ecca67f01c49fe5a68efd) I have written the beginning and the ending of the build process that the IDE uses for the example. Just in case it helps.
I have also included the beginning of the disassembly of the elf generated by the IDE [here](https://gist.github.com/Irvise/2499c2a6776d0040538ec57a32c113a6) and the [one generated by Ada](https://gist.github.com/Irvise/a1f6ed3caf57c6ad29e821407de48ada). Both were compiled with the debug option. At the beginning there are some noticeable differences (it is important to point out the IDE's example is using a simple `printf` which prints to serial, which may be the reason for the extra code added).

I am going to see in more detail how does the IDE upload the binary to the board, just in case I am missing something.

It may be better to mention this in #334 or #335 but since there is the "Metal library" for all the SiFive boards, maybe it is better to create an Ada wrapper around it? If it works, then, theoretically, all boards should be supported by the same "interface".

Fabien-Chouteau:
> I have also included the beginning of the disassembly of the elf generated by the IDE [here](https://gist.github.com/Irvise/2499c2a6776d0040538ec57a32c113a6) and the [one generated by Ada](https://gist.github.com/Irvise/a1f6ed3caf57c6ad29e821407de48ada)

Looks like I did a mistake on the flash base address: `0x20100000` instead of `0x20010000`.

> It may be better to mention this in #334 or #335 but since there is the "Metal library" for all the SiFive boards, maybe it is better to create an Ada wrapper around it? If it works, then, theoretically, all boards should be supported by the same "interface".

The goal of the Ada_Drivers_Library is to provide 100% Ada drivers for various platform. But it's true that this is not necessarily the most efficient way to use Ada on a board. Doing an Ada wrapper is a valid solution.

Fabien-Chouteau:
@Irvise can you try this new version?

Irvise:
@Fabien-Chouteau I am sorry I wrote the incorrect address in the issue. I tried the new version.

The generated binary is, as expected, "correct", the "boot" address is the correct one. Upon uploading it, it does nothing.

I have the impression that my changes mentioned in #334 were not correct. I think I tricked myself into thinking my changes worked because the program that was executing was doing a very similar thing to the example in Ada, so it may be more than just fixing the initial address. It seems that the "copy" method explained in the board's documentation may not be working as expected.

I tried to upload a know working .hex binary through this method and it failed. Therefore, I am going to first try to get a reliable upload mechanism I will get back in touch. Trying to trick the IDE by substituting the binaries has already failed, so I will try the freedom e sdk directly.

Irvise:
Good news @Fabien-Chouteau, I works!

I honestly cannot explain what is happening. To summarize what I have done is:
1. Built everything with with the Freedom Sdk. The examples work fine
2. Try to upload Ada's example. It failed, as in, the memory was left untouched (I did a memory dump to confirm this)
3. Built Ada's example in Production mode in a "desperate" attempt. It worked just fine!! I checked both ways of uploading the sketch (copypasting and with JLink)
4. Clean everything. Rebuild Ada's example in Debug mode. Expectation: it should fail. Reality, it now works for whatever reason

I don't know what has happened. But now it consistently works with both modes, Production and Debug. I will keep an eye on this and start fiddling with Ada. If I notice something out of the ordinary, I will let you know. However, as things are right now, I think this branch is ready to be merged.

As a side note, you may want to promote/comment/advertise the Ada driver in the official SiFive HiFive1's  forum. Some more people could be interested on it.

Regards and thank you,

Fernando Oleo Blanco

Fabien-Chouteau:
> I don't know what has happened. But now it consistently works with both modes, Production and Debug. I will keep an eye on this and start fiddling with Ada. If I notice something out of the ordinary, I will let you know. However, as things are right now, I think this branch is ready to be merged.

Thank you for the update @Irvise.


> As a side note, you may want to promote/comment/advertise the Ada driver in the official SiFive HiFive1's forum. Some more people could be interested on it.

Indeed, do you mind doing it for us?

PR #342, opened 2020-04-18
Add support for use of VL53L0x heightfinder in Crazyflie flow deck v1

simonjwright:
The Crazyflie flow deck v1 uses a VL53L0x "time of flight" sensor as a rangefinder to measure the aircraft’s height.

The VL53L0x can be set to generate an interrupt when a measurement is available, but according to [this schematic](https://wiki.bitcraze.io/_media/projects:crazyflie2:expansionboards:flow-deck_reve.pdf) the GPIO concerned (IO_2 on the left side expansion port) is multiplexed. The Crazyflie software handles this by setting the VL53L0x to continuous running, sampling when required. This change adds `Start_Continuous` to initiate continuous running.

Additionally, because the code to determine whether a measurement is actually available depends on the VL53L0x’s GPIO being configured to "generate an interrupt" when a new sample is available, preconditions are added to make sure that this is the case (if it isn’t, the software will loop indefinitely waiting for the sample).

### DIFF
--- a/components/src/range_sensor/VL53L0X/vl53l0x.adb
+++ b/components/src/range_sensor/VL53L0X/vl53l0x.adb
@@ -912,7 +912,7 @@ package body VL53L0X is
       end if;
 
       while not Range_Value_Available (This) loop
-         Delay_Milliseconds (1);
+         This.Timing.Delay_Milliseconds (1);
       end loop;
 
       return Read_Range_Millimeters (This);
--- a/components/src/range_sensor/VL53L0X/vl53l0x.adb
+++ b/components/src/range_sensor/VL53L0X/vl53l0x.adb
@@ -34,7 +34,6 @@
 --   COPYRIGHT(c) 2016 STMicroelectronics                                   --
 ------------------------------------------------------------------------------
 
-with Ada.Real_Time;
 with Ada.Unchecked_Conversion;
 
 with HAL;        use HAL;
@@ -906,7 +905,6 @@ package body VL53L0X is
      (This : VL53L0X_Ranging_Sensor) return HAL.UInt16
    is
       Status : Boolean;
-      use type Ada.Real_Time.Time;
    begin
       Start_Range_Single_Millimeters (This, Status);
       if not Status then
@@ -914,7 +912,7 @@ package body VL53L0X is
       end if;
 
       while not Range_Value_Available (This) loop
-         delay until Ada.Real_Time.Clock + Ada.Real_Time.Milliseconds (1);
+         Delay_Milliseconds (1);
       end loop;
 
       return Read_Range_Millimeters (This);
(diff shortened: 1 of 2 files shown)


simonjwright:
The build checks failed because the microbit zfp doesn’t support Ada.Real_Time. There ought to be a way to tell the CI not to try that!
I suppose I could make the offending subprogram into a generic which users have to instantiate with a procedure which is expected to either be null or to delay for a millisecond.
But other code in the library uses Ada.Real_Time?

simonjwright:
OK, just seen bosch_bno055.ads. I don’t know whether anyone is using the existing vl53l0x code, so will make just the offending subprogram (`Read_Range_Single_Millimeters`) generic, similarly.

Or I could revert this change and warn people about the spinloop? (my intended use doesn’t use this subprogram).

Fabien-Chouteau:
Hi @simonjwright ,

Thanks for the contribution.

> The build checks failed because the microbit zfp doesn’t support Ada.Real_Time. There ought to be a way to tell the CI not to try that!

The components drivers should be portable across any run-time profile. If you need timing features, the device should use an `HAL.Time.Any_Delays` implementation like here: https://github.com/AdaCore/Ada_Drivers_Library/blob/c958bb1d7fdf941b14b96eea61a78edb77216a10/components/src/screen/ST7735R/st7735r.ads#L53

If your run-time supports Ravenscar you can use this implementation: https://github.com/AdaCore/Ada_Drivers_Library/blob/c958bb1d7fdf941b14b96eea61a78edb77216a10/middleware/src/ravenscar-common/ravenscar_time.ads#L36

Otherwise it can be implemented by the BSP, like here for the micro:bit:
https://github.com/AdaCore/Ada_Drivers_Library/blob/c958bb1d7fdf941b14b96eea61a78edb77216a10/boards/MicroBit/src/microbit-time.ads#L64

simonjwright:
> OK, just seen bosch_bno055.ads. I don’t know whether anyone is using the existing vl53l0x code, so will make just the offending subprogram (`Read_Range_Single_Millimeters`) generic, similarly.

I hope this will be OK ... as noted, there is a precedent!

Fabien-Chouteau:
I don't like this approach because Ada's explicit generic instantiation make for an awkward API here.
I would rather prefer a `not null HAL.Time.Any_Delays` argument in this procedure if you don't want to add it to the type discriminants.

simonjwright:
> I don't like this approach because Ada's explicit generic instantiation make for an awkward API here.
> I would rather prefer a `not null HAL.Time.Any_Delays` argument in this procedure if you don't want to add it to the type discriminants.

Decided to add to the type discriminants; might want to add other delays later.

Fabien-Chouteau:
Thank you Simon!

PR #345, opened 2020-05-25
Migration to GNAT Community 2020

Fabien-Chouteau:


scripts/install_dependencies.py
```diff
@@ -41,24 +41,24 @@ def git_clone(repo_url, branch, dst, recursive=False):
     # Clone the repo
     returncode, stdout, stderr = run_program(
         ['git', 'clone', repo_url, dst] + extra_args)
-    print stdout
+    print(stdout)
```
setton, line 44: I would print the stdout only in the case the returncode is nonzero, below (same comment on line 56)

scripts/install_dependencies.py
```diff
@@ -71,7 +71,11 @@ def git_clone(repo_url, branch, dst, recursive=False):
               False,
               "bb-runtimes",
               False,
-              [sys.executable, ROOT_DIR + "/bb-runtimes/install.py", "--arch=arm-eabi"]),
+              [sys.executable, ROOT_DIR + "/bb-runtimes/install.py",
+               "--arch=arm-eabi",
+               "--prefix=" +
```
setton, line 76: (minor) why not use the .format() rather than string concatenation, for homogeneity with other places in the script?

.appveyor.yml
```diff
@@ -49,20 +49,32 @@ install:
...
 # Check that we can run the compiler (also display the version)
 - ps: arm-eabi-gnatls -v
 
+# Remove built-in run-times before installing fresh one. This is a workaround
+# for "Access is denied" in the bb-runtime Python script on appveyor.
+- ps: rm -Recurse -Force $env:INSTALL_DIR"\arm-elf\arm-eabi\lib\gnat\ravenscar*"
+- ps: rm -Recurse -Force $env:INSTALL_DIR"\arm-elf\arm-eabi\lib\gnat\zfp*"
+
 # Install optional dependencies
+- cmd: python --version
```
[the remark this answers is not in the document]
Fabien-Chouteau: I call it sanity-check. It doesn't cost anything and can save hours of pushing commits and waiting for appveyor to run.

scripts/install_dependencies.py
Fabien-Chouteau: Same as above, the more information the better.

scripts/install_dependencies.py
Fabien-Chouteau: I find it more readable in this case.

PR #346, opened 2020-05-27
Add board feature to initialize the Crazyflie external SPI.

simonjwright:
This unit already contains procedures (`Initialize_I2C_GPIO`, `Configure_I2C`) to do the hard work of initializing I2C.

There was no need to initialize SPI, because the basic Crazyflie doesn’t use it; however, SPI is used to communicate with the various [decks](https://www.bitcraze.io/documentation/system/platform/cf2-expansiondecks/).

This patch introduces a procedure to initialize the external SPI port.


boards/crazyflie/src/stm32-board.adb
```diff
@@ -1,6 +1,6 @@
 ------------------------------------------------------------------------------
 --                                                                          --
---                    Copyright (C) 2015, AdaCore                           --
+--                  Copyright (C) 2015, 2020 AdaCore                        --
```
Fabien-Chouteau, line 3: We use a dash `-` here between the two years: `2015-2020`.

boards/crazyflie/src/stm32-board.adb
```diff
@@ -173,4 +175,39 @@ package body STM32.Board is
       end if;
    end Configure_I2C;
 
+   procedure Initialize_EXT_SPI
```
Fabien-Chouteau, line 178: We add comment boxes above sub-program implementations:
```
------------------------
-- Initialize_EXT_SPI --
------------------------
```
You add a shortcut for that in GNATstudio, I use CTRL+B :)

Fabien-Chouteau:
Thank you @simonjwright , look like I have to fix the continuous integration before we can merge this PR.

boards/crazyflie/src/stm32-board.adb
simonjwright: I’ve done this, though I was intending to indicate that 2015, 2020 were the years in which changes were made, rather than 2015-2020, which in GCC would AFAIK mean that changes were made in all 6 years.

boards/crazyflie/src/stm32-board.adb
simonjwright: Done. Of course I can’t use GNATstudio, being on a Mac :-)

simonjwright:
I only realised after the fact that I should have included a procedure to initialize the Chip Select pin; done now

boards/crazyflie/src/stm32-board.adb
Fabien-Chouteau: As you said "The GPS from CE 2019 works just fine" :)

boards/crazyflie/src/stm32-board.adb
Fabien-Chouteau: This is our policy, we also put the range.

Fabien-Chouteau:
Why don't you use the CS pin as an argument to the first procedure and merge the two?

simonjwright:
> Why don't you use the CS pin as an argument to the first procedure and merge the two?

Hadn’t thought of that - but I think the reason is, there could be more than one peripheral, sharing the SPI but each with its own CS line.

Fabien-Chouteau:
Hi @simonjwright , do you mind re-basing your patches on `master`? I have merged the Community 2020 support.

Thanks,

simonjwright:
I think I didn’t do the rebasing as neatly as I could, sorry

Fabien-Chouteau:
You can use this command:
```
git pull --rebase --autostash https://github.com/AdaCore/Ada_Drivers_Library master
```

simonjwright:
Replaced by #349 .

PR #347, opened 2020-05-27
Add PMW3901 (flow) sensor support.

simonjwright:
The [PMW3901 sensor from PixArt Imaging](https://www.pixart.com/products-comparison/16/Optical_Motion_Tracking) detects lateral motion of the sensor, using optical imaging. This can be used for drone stabilization.

~This patch includes a `separate` function to check whether an SPI port is enabled. This is used in a precondition.
It might be thought reasonable to include such a feature in the standard `STM32.Device` unit; on the other hand , it’ll be pretty obvious if the device isn’t enabled, since nothing will happen, so maybe the precondition isn’t a good idea. It could be removed.~


simonjwright:
OK, I see what I did wrong there. Will rethink.

components/src/motion/pmw3901/pmw3901.ads
```diff
@@ -0,0 +1,109 @@
...
+
+with STM32.GPIO;
+with STM32.SPI;
+with HAL.Time;
+with Interfaces;
+
+package PMW3901 is
+
+   type PMW3901_Flow_Sensor
+     (Port   : not null access STM32.SPI.SPI_Port;
```
Fabien-Chouteau, line 44: To be portable, components cannot depend on specific MCU drivers.
You have to use `HAL.SPI` and `HAL.GPIO` here.

components/src/motion/pmw3901/pmw3901.adb
```diff
@@ -0,0 +1,310 @@
...
+   function Read_Motion  (This : in out PMW3901_Flow_Sensor) return Motion is
+      subtype Buffer is STM32.SPI.UInt8_Buffer (1 .. 12);
+      function Convert is new Ada.Unchecked_Conversion (Buffer, Motion);
+      Dummy : HAL.UInt8;
+      Buff : Buffer := (others => 0);
+   begin
+      This.CS.Drive (False);
+      This.Port.Transmit_Receive (Outgoing => Reg_Motion_Burst,
+                                 Incoming => Dummy);
+      pragma Warnings (Off, "referenced before it has a value");
```
Fabien-Chouteau, line 264: This is a pretty important warning you are silencing. Maybe you can have a different buffer incoming and outgoing.

components/src/motion/pmw3901/pmw3901.adb
```diff
@@ -0,0 +1,310 @@
...
+      use type HAL.UInt8;
+   begin
+      return M.Motion_Occurred
+        and not (M.S_Qual < 19 and M.Shutter_Upper = 16#1f#);
+   end Is_Valid;
+
+   function Is_Clocked (This : STM32.SPI.SPI_Port'Class) return Boolean;
+   --  It'd be good if this sort of function was declared in STM32.Device
+
+   function Is_Clocked (This : PMW3901_Flow_Sensor) return Boolean
```
Fabien-Chouteau, line 69: Subprogram comment box are missing (see #346).

components/src/motion/pmw3901/pmw3901.ads
```diff
@@ -0,0 +1,109 @@
...
+
+   function Is_Valid (M : Motion) return Boolean;
+
+   function Is_Clocked (This : PMW3901_Flow_Sensor) return Boolean;
+
+   function Is_Initialized (This : PMW3901_Flow_Sensor) return Boolean;
+
+   procedure Initialize (This : in out PMW3901_Flow_Sensor)
+   with
+     Pre => Is_Clocked (This);
```
Fabien-Chouteau, line 89: Is that to specify that the SPI should be on?

We don't do that kind of checks on other drivers, the intention is good but there is not much added value here.
For instance the SPI could be on but not configured.

Fabien-Chouteau:
Thank you @simonjwright this is a cool sensor.

components/src/motion/pmw3901/pmw3901.ads
simonjwright: Yes, that was what I’d spotted!

components/src/motion/pmw3901/pmw3901.adb
simonjwright: For some reason, all the code I’ve seen insisted on using this form (transmit  request, receive dummy, transmit dummy, receive result of request). Why, I don’t know, because doing it as you would have expected (transmit the request, receive the result) works just fine; new code doesn’t provoke this warning.

components/src/motion/pmw3901/pmw3901.adb
simonjwright: OK

components/src/motion/pmw3901/pmw3901.ads
simonjwright: I’ve changed it to `not Initialized`, so I don’t do it twice.
Also, very difficult to do the 'is clocked' check in a HAL-compatible way, I think, so I’ve removed it.

Fabien-Chouteau: The solution for an HAL-compatible `Is_Clocked` would be to add it to the interface.
But the way HAL is used right now is that components expect an enabled and configured SPI so they don't need to check if it's on or not, configured or not.

Fabien-Chouteau:
Hi @simonjwright , do you mind re-basing your patches on `master`? I have merged the Community 2020 support.

Thanks,

simonjwright:
Successfully rebased, hope OK

Fabien-Chouteau:
Also needs a rebase from `master`:
```
git pull --rebase --autostash https://github.com/AdaCore/Ada_Drivers_Library master
```

simonjwright:
Look, I’m sorry, this has ended up with **16** commits! I’d be quite happy to cancel this merge request and start again.
I expect the same thing will happen with #346.

Fabien-Chouteau:
Looks like it is because your original commit (https://github.com/AdaCore/Ada_Drivers_Library/pull/347/commits/89c54e73732151bd6c2b4eb1becb36a4e04c6344) is still in the branch and before mines.

simonjwright:
Added a cleaned-up pull request #348.

(Thinking back, I don’t see why it was necessary to rebase? the 2020 changes and these didn’t appear to impact each other)

PR #349, opened 2020-06-03
Add board feature to initialize the Crazyflie external SPI.

simonjwright:
This unit already contains procedures (`Initialize_I2C_GPIO`, `Configure_I2C`) to do the hard work of initializing I2C.

There was no need to initialize SPI, because the basic Crazyflie doesn’t use it; however, SPI is used to communicate with the various [decks](https://www.bitcraze.io/documentation/system/platform/cf2-expansiondecks/).

This patch introduces a procedure to initialize the external SPI port, and another to configure the required Chip Select pin.

Replaces pull request #346.


Fabien-Chouteau:
Thank you Simon :)

PR #352, opened 2020-06-16
Project wizard: don't optimize semihosting.adb

simonjwright:
This problem didn’t show up with GNAT CE 2019, but it does with FSF GCC 9.1.0 & 10.1.0: the output of semihosting is garbled if semihosting.adb is compiled with -O3, OK if compiled with -O0.

What should have been
```
One_Wire.Get_Info:
  13  46  71  31  0  0  0  135
Flow & ToF both initialized
Flow_Sensor_Task started
ToF_Sensor_Task started
```
came out as
```
??

P?
?? 

q
```

Setup: GDB 8.3 & 9.1 rsp, SEGGER J-Link Edu, SEGGER J-Link GDB Server V6.42 Command Line Version, macOS Catalina

### DIFF
--- a/arch/ARM/cortex_m/src/semihosting.adb
+++ b/arch/ARM/cortex_m/src/semihosting.adb
@@ -1,6 +1,6 @@
 ------------------------------------------------------------------------------
 --                                                                          --
---                    Copyright (C) 2016, AdaCore                           --
+--                  Copyright (C) 2016-2020, AdaCore                        --
 --                                                                          --
 --  Redistribution and use in source and binary forms, with or without      --
 --  modification, are permitted provided that the following conditions are  --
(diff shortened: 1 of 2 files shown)


Fabien-Chouteau:
Hi @simonjwright ,

Do you know what is optimized exactly?

It would be best to fix this in the source code rather than the project file. You change will break builds for RISC-V which doesn't have `semihosting.adb`.

simonjwright:
Point taken.

Commit cf3d488 works for me. I did look at the ASM but I really didn’t know what to look for; guessed that forcing copying byte-by-byte might do the trick.

The title of this pull request is now OBE; and it includes a commit that you wouldn’t want. Assuming you’re OK with the change, how should I proceed?

Fabien-Chouteau:
Thanks @simonjwright , the last patch looks good.

You can add a revert commit on this PR for e7884d5 or start from a new PR.

Either way please rebase on master since I just merged a change in the continuous integration scripts.

simonjwright:
> You change will break builds for RISC-V which doesn't have `semihosting.adb`.

I don’t believe this is true. [This GPR](https://github.com/simonjwright/cortex-gnat-rts/blob/master/stm32f4/build_runtime.gpr#L89) from my Cortex GNAT RTS specifies `Switches` for at least 3 files that aren’t present in that RTS.

simonjwright:
Closed in favour of #353

PR #353, opened 2020-06-17
Reduce optimization opportunities in Semihosting.

simonjwright:
This problem didn’t show up with GNAT CE 2019, but it does with FSF GCC 9.1.0 & 10.1.0: the output of semihosting is garbled if semihosting.adb is compiled with -O3, OK if compiled with -O0.

What should have been
```
One_Wire.Get_Info:
  13  46  71  31  0  0  0  135
Flow & ToF both initialized
Flow_Sensor_Task started
ToF_Sensor_Task started
```
came out as
```
??

P?
?? 

q
```

Setup: GDB 8.3 & 9.1 rsp, SEGGER J-Link Edu, SEGGER J-Link GDB Server V6.42 Command Line Version, macOS Catalina

### DIFF
--- a/arch/ARM/cortex_m/src/semihosting.adb
+++ b/arch/ARM/cortex_m/src/semihosting.adb
@@ -1,6 +1,6 @@
 ------------------------------------------------------------------------------
 --                                                                          --
---                    Copyright (C) 2016, AdaCore                           --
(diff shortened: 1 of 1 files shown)


Fabien-Chouteau:
Thank you @simonjwright

PR #370, opened 2020-09-18
Add first draft of FE310.Performance_Monitor package.

### LINKED ISSUE #365: Access to FE310-specific performance counters
Following PR #362, we can now implement support for the micro-architectural performance counters implemented by the FE310 in the HiFive1. Only 2 such counters are implemented (mhpmcounter3 and 4), controlled by mhpmevent3 and 4 control registers, with specific meaning defined bu the FE310 Reference Manual.

rod-chapman:
Opening Pull Request.

Closes #365

### DIFF
--- a/arch/RISC-V/SiFive/drivers/fe310-performance_monitor.adb
+++ b/arch/RISC-V/SiFive/drivers/fe310-performance_monitor.adb
@@ -1,0 +1,151 @@
+------------------------------------------------------------------------------
+--                                                                          --
+--                       Copyright (C) 2020, AdaCore                        --
+--                                                                          --
+--  Redistribution and use in source and binary forms, with or without      --
+--  modification, are permitted provided that the following conditions are  --
+--  met:                                                                    --
+--     1. Redistributions of source code must retain the above copyright    --
+--        notice, this list of conditions and the following disclaimer.     --
+--     2. Redistributions in binary form must reproduce the above copyright --
+--        notice, this list of conditions and the following disclaimer in   --
+--        the documentation and/or other materials provided with the        --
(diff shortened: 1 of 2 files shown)


rod-chapman:
I am having trouble getting sensible values back from mhpmcounter3 - the most-significant 32 bits seem to be corrupted in some way and don't make any sense. The FE310 manual says something about the counter only really being 40 bits, so I tried to mask off the top-most 24 bits (i.e. doing andi t3, t3, 255 after the csrr t3, mhpmcounter3h instructions), but that didn't help...

Does anyone at AdaCore have way to test this?

Fabien-Chouteau:
> Does anyone at AdaCore have way to test this?

I have a HiFive1 revA here, but did you try with the C API first?
Are there examples provided in the SDK?

rod-chapman:
I tried using my own API to call to assembly code, with the same results. If I reset the counter to 0 at the start of the test, then only read the lower 32 bits, then it seems to be OK, but the upper 32 bits are all wrong...

Fabien-Chouteau:
Maybe you can ask on the SiFive forums.

Fabien-Chouteau:
@rod-chapman did you get more info on this?

The code looks good I am ok to merge this.

rod-chapman:
OK - please go ahead and merge.

rod-chapman:
I didn't get the 64-bit counters (like mhpmcounter3) to work reliably - I was still getting back garbage values for the 32 MSBs. At the moment, I am not using them, so this is not blocking me.

Fabien-Chouteau:
Thank yo @rod-chapman

PR #372, opened 2020-09-22
Fix LED turn on/off logic

maksumic:
Tested on Arduino Nano 33 BLE Sense


Fabien-Chouteau:
Hello @maksumic ,

On the board I have the logic is good. Also looking at the schematics, the GPIO are sinking current for the LED so the current code is correct.

What board did you test this on?

PR #387, opened 2021-06-23
Add Cortex_M.Hints, which provides access to hint instructions

JeremyGrosser:
Send_Event, Wait_For_Event, Wait_For_Interrupt, and Yield are defined as "hint instructions" in the [ARMv6-M Architecture Reference Manual](https://developer.arm.com/documentation/ddi0419/e). This package just conveniently wraps them up as procedures, very similar to the Memory_Barriers package in the same directory.

When testing this package, I discovered that GNAT's `Inline` is really just a suggestion. Unless `-fomit-frame-pointer` is added to the compiler flags, five additional instructions are added to each of these calls to save and restore the link register. Here is the disassembled output of Wait_For_Interrupt annotated with source.
```
   procedure Wait_For_Interrupt is
     908:       b580            push    {r7, lr}
     90a:       af00            add     r7, sp, #0
      pragma Suppress (All_Checks);
   begin
      Asm ("wfi", Volatile => True);
     90c:       bf30            wfi
   end Wait_For_Interrupt;
     90e:       46bd            mov     sp, r7
     910:       bd80            pop     {r7, pc}
     912:       46c0            nop                     ; (mov r8, r8)
```

Even with `-fomit-frame-pointer`, this function still requires a `bl` branch with link to the `wfi` instruction, followed by a `bx lr` afterward.

Is there any aspect or pragma that can be used to cause GNAT to truly inline these instructions without branching or manipulating the stack?

### DIFF
--- a/arch/ARM/cortex_m/src/cortex_m-hints.adb
+++ b/arch/ARM/cortex_m/src/cortex_m-hints.adb
@@ -1,0 +1,71 @@
+------------------------------------------------------------------------------
+--                                                                          --
+--                    Copyright (C) 2021, AdaCore                           --
+--                                                                          --
+--  Redistribution and use in source and binary forms, with or without      --
+--  modification, are permitted provided that the following conditions are  --
+--  met:                                                                    --
+--     1. Redistributions of source code must retain the above copyright    --
+--        notice, this list of conditions and the following disclaimer.     --
+--     2. Redistributions in binary form must reproduce the above copyright --
+--        notice, this list of conditions and the following disclaimer in   --
+--        the documentation and/or other materials provided with the        --
+--        distribution.                                                     --
+--     3. Neither the name of the copyright holder nor the names of its     --
+--        contributors may be used to endorse or promote products derived   --
+--        from this software without specific prior written permission.     --
+--                                                                          --
+--   THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS    --
+--   "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT      --
+--   LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR  --
+--   A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT   --
+--   HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, --
+--   SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT       --
+--   LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,  --
+--   DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY  --
+--   THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT    --
+--   (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE  --
+--   OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.   --
+--                                                                          --
+------------------------------------------------------------------------------
+with System.Machine_Code; use System.Machine_Code;
+
+package body Cortex_M.Hints is
+
+   -----------
+   -- Yield --
+   -----------
+
+   procedure Yield is
+   begin
+      Asm ("yield", Volatile => True);
+   end Yield;
+
+   ----------------
+   -- Send_Event --
+   ----------------
+
+   procedure Send_Event is
+   begin
+      Asm ("sev", Volatile => True);
+   end Send_Event;
+
+   --------------------
+   -- Wait_For_Event --
+   --------------------
+
+   procedure Wait_For_Event is
+   begin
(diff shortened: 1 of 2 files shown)


arch/ARM/cortex_m/src/cortex_m-hints.adb
```diff
@@ -0,0 +1,59 @@
...
+--   (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE  --
+--   OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.   --
+--                                                                          --
+------------------------------------------------------------------------------
+with System.Machine_Code; use System.Machine_Code;
+
+package body Cortex_M.Hints is
+
+   procedure Yield is
+      pragma Suppress (All_Checks);
```
Fabien-Chouteau, line 36: Why do you add `pragma Suppress` here?

JeremyGrosser:
Upon further investigation, enabling Link Time Optimization by adding `-flto` to both the compiler and linker flags does indeed inline this without branching.

arch/ARM/cortex_m/src/cortex_m-hints.adb
```diff
@@ -0,0 +1,59 @@
...
+--   THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT    --
+--   (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE  --
+--   OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.   --
+--                                                                          --
+------------------------------------------------------------------------------
+with System.Machine_Code; use System.Machine_Code;
+
+package body Cortex_M.Hints is
+
+   procedure Yield is
```
Fabien-Chouteau, line 35: We add a "comment box" for sub-program implementation in the body, e.g. : https://github.com/AdaCore/Ada_Drivers_Library/blob/0ccb49fb9ff53d0cf621b034a234fa3479576136/arch/ARM/cortex_m/src/cortex_m-dwt.adb#L38

You can define a shortcut to make the comment boxes in GNATstudio.

Fabien-Chouteau (changes requested):
Thank you @JeremyGrosser , I have a couple of comments on the coding style mostly.

Fabien-Chouteau:
I think the compiler should inline as soon as you enable some -O2 or -O3 optimization at least.
There is also `-gnatn` to enable inlining.

And you can give a hints to the compiler with `pragma Inline (Yield);` at the end of the specification for instance.

arch/ARM/cortex_m/src/cortex_m-hints.adb
JeremyGrosser: I was copying from Memory_Barriers. It does appear that this is unnecessary. I'll remove it.

arch/ARM/cortex_m/src/cortex_m-hints.adb
JeremyGrosser: Right, I'll add those.

JeremyGrosser:
> I think the compiler should inline as soon as you enable some -O2 or -O3 optimization at least.
> There is also `-gnatn` to enable inlining.
> 
> And you can give a hints to the compiler with `pragma Inline (Yield);` at the end of the specification for instance.

I've already got `with Inline` on the procedures, so `-gnatn` is redundant. Compiling with `-O3` does remove the link register manipulation, but not the branching. The only way I've found to do that is to enable LTO.

JeremyGrosser:
> I think the compiler should inline as soon as you enable some -O2 or -O3 optimization at least.
> There is also `-gnatn` to enable inlining.
> 
> And you can give a hints to the compiler with `pragma Inline (Yield);` at the end of the specification for instance.

I already have `with Inline` on all of the procedures in the specification, the Ada 2012 reference says this is preferred over the pragma now.

It looks like adding `-O3` removes the link register manipulation and `-gnatn` gets rid of the branching. I would've expected that `with Inline` or `-O3` would've implicitly enabled inlining, but apparently not!

The GNAT user guide makes this set of conditions clear here https://gcc.gnu.org/onlinedocs/gcc-10.2.0/gnat_ugn/Inlining-of-Subprograms.html#g_t10f

arch/ARM/cortex_m/src/cortex_m-hints.ads
```diff
@@ -0,0 +1,70 @@
...
+--
+--  Source:
+--
+--    ARMv6-M Architecture Reference Manual
+--    A6.6 Hint Instructions
+--    https://developer.arm.com/documentation/ddi0419/e
+
+package Cortex_M.Hints is
+
+   --  A6.7.57 SEV
```
Fabien-Chouteau, line 43: I am sorry I forgot to hit "send" on the last comment.

We follow the GNAT coding style where comments go after the sub-program definition: https://github.com/AdaCore/Ada_Drivers_Library/blob/0ccb49fb9ff53d0cf621b034a234fa3479576136/arch/ARM/cortex_m/src/cortex_m-dwt.ads#L60

Fabien-Chouteau (changes requested):

Fabien-Chouteau (approved):

Fabien-Chouteau:
@JeremyGrosser thanks for the contribution. The builds are failing because of another issue in the scripts (see #388).
I'll ask you to rebase once this is fixed.

Fabien-Chouteau:
@JeremyGrosser if you can rebase on the `master` branch that would be great :)

Fabien-Chouteau (approved):

Fabien-Chouteau:
Thank you for your contribution @JeremyGrosser, and sorry for the issue with the build script.

PR #389, opened 2021-07-19
SGTL5000: fix I2C_Write and a few config procedures

Fabien-Chouteau:
The current implementation worked fine on the STM32F4, but it could be
because the I2C Mem_Write is not correctly implemented for this
microcontroller.


pat-rogers (approved):

Fabien-Chouteau:
Thanks Pat,

Issue #340, opened 2020-04-14
Handling of drivers referencing non-existant hardware features in svd


Hendrik410:
Hello,

I am currently implementing the STM32F411RE Nucleo board.
I have downloaded the svd files from ST and generated the svd ada files.

I've then copied the device folder from another device and started modifying. 
While doing so, I commented all the periphirals out, that the STM32F411 doesn't have (The extra timers, the extra usarts, the dac).

I've also created the necessary entries and runtime projects for the board.

When I open the duplicated example project, the compiler copmlains about the missing references of from the stm drivers. An example is seen here:
![image](https://user-images.githubusercontent.com/11255608/79235087-718eb600-7e6b-11ea-9d96-7b1d051c5170.png)

Now if I understand it correctly, the drivers are common for all STM32 devices.

My question is, how should missing peripherals be handled?

Fabien-Chouteau:
Hello @Hendrik410 ,

> I've also created the necessary entries and runtime projects for the board.

Did you add the board and device definition in the [config scripts](https://github.com/AdaCore/Ada_Drivers_Library/tree/master/scripts/config)?

> Now if I understand it correctly, the drivers are common for all STM32 devices.

They are at least common to all the STM32 devices that we support so far. But since we do not know the hundreds of STM32 devices available (and the ones to come), 

> My question is, how should missing peripherals be handled?

There is no easy answer I am afraid. Adding to discussion @pat-rogers who wrote this package.

Hendrik410:
Hi @Fabien-Chouteau ,

yes, I've used the config scripts.

I'm not to firm in Ada, I know that in C, compiler dirictives would be used. 
Is there a similar construct in Ada?

Fabien-Chouteau:
No, Ada doesn't have the same feature, which is a good thing in most cases :)

@pat-rogers what is your opinion on this?

pat-rogers:
>     Now if I understand it correctly, the drivers are common for all
>     STM32 devices.
> 
> They are at least common to all the STM32 devices that we support so 
> far. But since we do not know the hundreds of STM32 devices available 
> (and the ones to come),
> 
>     My question is, how should missing peripherals be handled?
> 
> There is no easy answer I am afraid. Adding to discussion @pat-rogers 
> <https://github.com/pat-rogers> who wrote this package.


When I wrote this package I admit thinking to myself that these "timer 
classifier" functions needed to go in some device-specific location. A 
[...]

PR #391, opened 2021-10-31
Adding Nucleo_f401re with examples

### LINKED ISSUE #313: Support for STM32401RE board
Hi,

will STM32401RE be supported? It seems there's no driver for this board. 
How can I contribute in order to support this board?

FredPraca:
Adding files for Nucleo_f401re in order to fix #313


FredPraca:
For stm32-adc, STM32F401 doesn't support multi-mode.
For timers, TIM6 and TIM7 don't exist and there are no basic timers for STM32F401 or STM32F411.
For PWM, the fact that several timers are missing in these series has an impact on _has_APB1__Frequency_.

FredPraca:
After reading #340, it seems that redefining these files was a bad idea but that's the only one I got to make it work for both Nucleo F401 and Nucleo F411 in a simple way.

PR #394, opened 2021-11-27
Components: add I2C, SPI drivers for LPS25H barometric sensor.

simonjwright:
These drivers don’t support FIFO operation; just call Get_Data to retrieve the latest value.

The altitude above sea level value returned by Get_Data assumes a (very) simplified model.


Fabien-Chouteau (approved):

Fabien-Chouteau:
Thanks for your contribution @simonjwright

PR #400, opened 2022-02-23
Add pragma Preelaborate to all HAL packages.

### LINKED ISSUE #399: Make HAL packages preelaborable
I'd like to add `pragma Preelaborate` to a few packages in rp2040_hal, but I can't do this as long as there's a dependency on the HAL packages, which are not preelaborated. Could we add Preelaborate to the HAL packages in ADL?

JeremyGrosser:
Fixes #399. I used the pragma rather than an aspect just because the top level HAL package uses a pragma. Let me know if you'd prefer `with Preelaborate`.

### DIFF
--- a/hal/src/hal-audio.ads
+++ b/hal/src/hal-audio.ads
@@ -32,6 +32,7 @@
 with Interfaces; use Interfaces;
 
 package HAL.Audio is
+   pragma Preelaborate;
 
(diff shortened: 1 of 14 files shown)


Fabien-Chouteau (approved):

Fabien-Chouteau:
Thanks @JeremyGrosser

PR #404, opened 2022-05-01
Comp new pcf8574

RREE:
The PCF8574 is an old and simple io-expander from the 90ies without data direction register nor latch. GPIO_Pin cannot be implemented (I think). It is widely used in small LCD displays.

### DIFF
--- a/components/src/io_expander/pcf8574/pcf8574.adb
+++ b/components/src/io_expander/pcf8574/pcf8574.adb
@@ -1,0 +1,51 @@
+--
+--  Copyright 2022 (C) Rolf Ebert
+--
+--  SPDX-License-Identifier: BSD-3-Clause
+--
+
+
+package body PCF8574 is
+
+   -----------------
+   --  Configure  --
+   -----------------
+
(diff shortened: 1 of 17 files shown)


components/src/io_expander/pcf8574/pcf8574.adb
```diff
@@ -0,0 +1,51 @@
...
+   -----------
+   --  Get  --
+   -----------
+
+   function Get (This : PCF8574_Module) return UInt8
+   is
+      Val    : I2C_Data (1 .. 1);
+      Status : I2C_Status;
+   begin
+      This.Port.Receive (This.Addr, Val, Status);
```
Fabien-Chouteau, line 31: `This.Port` is potentially `null`, you should check that before.
Same thing for the `Set` procedure.

Fabien-Chouteau (changes requested):
Thanks for the contribution @RREE.

I have a couple comments.

Fabien-Chouteau:
See #405

PR #414, opened 2022-08-22
Avoid unnecessary conversions when utilising the Interfaces package on the compatible HAL types.

kevlar700:


Fabien-Chouteau (commented):
Hello @kevlar700,

The types used to be derived from Interfaces but we changed that long ago. See: https://github.com/AdaCore/Ada_Drivers_Library/commit/0a804815ae8e559fe09150c6d06f94832398b14f

You mention unnecessary explicit conversions to `Interfaces.Unsigned_*` types, but with this change you have extra `with` of the `Interfaces` package and a discrepancy between `UInt8`, `UInt16`, `UInt32` and the other types defined in `HAL`. 

I wonder if should actually remove the dependency on `Interfaces` completely and define `UInt8`, `UInt16`, `UInt32` like the other types.

kevlar700:
Why wouldn't you want to use Interfaces functions like shift without conversions?

Fabien-Chouteau:
The `Shift_Left` `Shift_Right` functions are available with the current code, without conversions.

Fabien-Chouteau:
These functions are intrinsics, see definition in `Interfaces`: https://github.com/Fabien-Chouteau/bare_runtime/blob/bd95e2b7d0911d72370e2f9abcd7c11d77690a75/src/interfac.ads#L89

The `UInt8`, `UInt16` and `UInt32` type in `HAL` inherit them from the `Intrerfaces.Unsigned_` types.

We should actually declare those functions for all the other `UInt*` types here.

pat-rogers:
Right, the shift/rotate ops are inherited when deriving from those types declared in Interfaces,, and are not otherwise defined for user-defined modular types. We could explicitly declare the functions, as intrinsics, or we could apply the GNAT-defined pragma Provide_Shift_Operators.

PR #415, opened 2022-11-05
add new IO expander MCP23017 (2x8bits)

RREE:
The MCP23017 is similar to the existing MCP23008, it has two ports of 8 pins.
Using the latest 12.2 compiler I was nudged to use square brackets for arrays. That might be problem if you want to use it with an older compiler


Fabien-Chouteau (approved):

Fabien-Chouteau:
Thanks @RREE

Issue #393, opened 2021-11-22
mma8653 initialization fails on MicroBit


rveenker:
The accelerometer example for the Micro:Bit fails to initialize. (Using Micro:Bit v1.3B, firmware 0249)
Running a Python example however works fine and once initialized, the Ada accelerometer example will also work.
But after a reboot (power off/on), the Ada accelerometer fails again, and I keep getting the same result values.

By attaching a logic analyzer (Smartscope) to the i2c bus, I noticed a small difference in the way the register values are transmitted to the mma8653. 
I have attached two screenshots of the initialization procedure. It appears that the Ada version emits extra address fields for the register values, whereas the Python version uses a single address field for both register id and its value. 

Ada version:
![MicroBit_Accelerator_Ada](https://user-images.githubusercontent.com/10026531/142933095-2110a65b-3306-4ea5-aa77-c20603c76709.jpg)

Python version:
![MicroBit_Accelerator_Python](https://user-images.githubusercontent.com/10026531/142933112-49accf99-df1a-436d-9d52-ed79229b052b.jpg)

Fabien-Chouteau:
Hi Rob,
The I2C protocol can be used in different ways. The Python version doesn't send stop/start tokens between address and data access, but that doesn't necessarily explain the problem.

What error do you see?

rveenker:
Hello Fabien,
What I see is that the values from the mma8653 do not change at all and sometimes there are no values at all. 
Only after running another program that also uses the mma8653, I get results that make sense.
So my guess was that the problem is caused by incorrect or incomplete initialization.
For testing, I modified the accelerometer so that it sends the exact same register values as the python program, but this did not solve the issue. 
Do you have suggestions on how to proceed? Thanks!

Fabien-Chouteau:
> For testing, I modified the accelerometer so that it sends the exact same register values as the python program, but this did not solve the issue.

That's what I would have recommended. So if it doesn't work I don't know what to say...

simonjwright:
I couldn't get the accelerometer example to work (it was constantly returning the same values). I tried the _mu_ MicroPython development environment as recommended [here](https://microbit-micropython.readthedocs.io/en/v1.0.1/tutorials/introduction.html), and that didn't work either. So I tried restoring the standard firmware instead of the Segger JLink version - success!

Need to check out pyOCD.

rveenker:
[...]

PR #417, opened 2022-12-02
Correct the transmit protocol for Nordic TWI.

simonjwright:
For a write, the register address and the data must be sent in one operation, as specified in the nRF51 Series Reference Manual version 3.0.

This corrects issue #393.

### DIFF
--- a/arch/ARM/Nordic/drivers/nrf_common/nrf-twi.adb
+++ b/arch/ARM/Nordic/drivers/nrf_common/nrf-twi.adb
@@ -206,7 +206,6 @@ package body nRF.TWI is
       --  Set Address
       This.Periph.ADDRESS.ADDRESS := UInt7 (Addr / 2);
 
-
       if Data'Length = 1 then
          --  Only one byte to receive so we stop at the next one
          This.Periph.SHORTS.BB_STOP := Enabled;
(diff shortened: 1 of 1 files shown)


arch/ARM/Nordic/drivers/nrf_common/nrf-twi.adb
```diff
@@ -266,34 +271,29 @@ package body nRF.TWI is
...
 
-      This.Do_Stop_Sequence := False;
+      This.Do_Stop_Sequence := True;
 
       case Mem_Addr_Size is
          when Memory_Size_8b =>
             This.Master_Transmit (Addr    => Addr,
-                                  Data    => (0 => UInt8 (Mem_Addr)),
+                                  Data    =>
+                                    (0 => UInt8 (Mem_Addr)) & Data,
```
Fabien-Chouteau, line 280: One issue here is that if you send a big `Data` array (e.g. a frame buffer for a screen), you have a pretty high risk of  overflowing the stack with this concatenation operation.

simonjwright: True, but (a) is that likely to happen with I2C devices? (b) how could it be addressed, anyway? (c) could we rely on the device being OK with doing the transfer in multiple chunks? (d) what would the chunk size limit be?

I’d imagine that frame buffers would be sent in parts anyway.

We could offer a different interface for chunkable transfers, but you’d still need to concatenate the address and the data given the way TWI works.

Perhaps SUSPEND might offer a solution?

simonjwright: OK, I was wrong about this. The second part of my changes was probably a remnant of experimental changes to try to get _some_ handle on what was/wasn’t working; reverted.

Fabien-Chouteau (approved):

PR #427, opened 2024-02-27
Provide control over -fcallgraph-info in project wizard.

simonjwright:
In [GCC PR 104342](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=104342) we can see that, if checking out a non-release compiler (e.g. GCC 14.0.1), having `-gnata` and `-fcallgraph-info=su` will cause a crash (in stm32 code, anyway).

`-gnata` is used when Build is Debug.

This patch amends the Project Wizard so you can specify `CALLGRAPHS=Disabled`.


pat-rogers (approved):

pat-rogers:
Thanks Simon.

PR #437, opened 2024-11-17
Wait for STM32 timer output to be enabled before returning from enable procedure.

liampwll:
RM0386 Rev 6 22.3.12 notes that there may be a delay before the BDTR.MOE bit reads as 1 after writing to it. The RM says that only a single instruction is needed, but testing has shown that this is incorrect.

With the previous code, calling Enable_Main_Output and then immediately calling a procedure to set a different part of the register could result in the MOE bit being inadvertently set low.

### DIFF
--- a/arch/ARM/STM32/drivers/stm32-timers.adb
+++ b/arch/ARM/STM32/drivers/stm32-timers.adb
@@ -184,6 +184,10 @@ package body STM32.Timers is
    procedure Enable_Main_Output (This : in out Timer) is
    begin
       This.BDTR.Main_Output_Enabled := True;
+
+      loop
+         exit when Main_Output_Enabled (This);
+      end loop;
    end Enable_Main_Output;
 
    -------------------------


liampwll:
The same also applies to BKE and BKP as noted by the previous paragraph in the reference manual:

> When the BKEx and BKPx bits are written, a delay of 1 APB clock cycle is applied before the writing is effective. Consequently, it is necessary to wait 1 APB clock period to correctly read back the bit after the write operation.

It's very likely that this is causing issues in the following code in stm32-timers.adb, but I haven't tested this since we don't use it:

```
      This.BDTR.Automatic_Output_Enabled      := Automatic_Output_Enabled;
      This.BDTR.Break_Polarity                := Break_Polarity;
      This.BDTR.Break_Enable                  := Break_Enabled;
      This.BDTR.Off_State_Selection_Run_Mode  := Off_State_Selection_Run_Mode;
```

pat-rogers (approved):

PR #448, opened 2025-08-30
Extensively refine code for these three serial I/O examples

pat-rogers:
README.md
- Add new readme indicating that this host program is only intended for the
stream-based demo, and explaining how to use it

host.adb
- Add much more comments in the header block explaining how to use this program and
why it is needed
- Parameterize the host port number instead of hard-coding it

demo_serial_port_blocking.adb
- Use normalized procedure names for Send and Receive, for consistency with
other versions
- Add comments re: baud rate and other settings required

demo_serial_port_nonblocking.adb
- Better name for local procedure
- Add many more comments re: settings

demo_serial_port_streaming.adb
- Replace source code in comment header block with location of the actual
code and project file.
- Add comment re: serial port settings

peripherals_blocking.ads
- Peripheral should be a constant
- COM now references the USART designated by Peripheral

peripherals_nonblocking.ads
- Same as for peripherals_blocking.ads

peripherals_streaming.ads
- Same as for peripherals_blocking.ads

serial_io-blocking.ads
- Type Serial_Port now designates a USART rather than type Peripheral
- No longer provide call-through routines for the Serial_IO package's
Initialize_Hardware and Configure since client can easily call them. Better
separation of concerns as a result.
- Change signatures of internal routines to be access-to-USART

serial_io-blocking.adb
- Internal simplifications reflecting change to access discriminant now designating type USART
- Remove code for Initialize_Hardware and Configure, per removal from package spec

serial_io-nonblocking.ads
- Type Serial_Port now designates a USART rather than type Peripheral
- No longer provide call-through routines for Serial_IO packages
Initialize_Hardware and Configure since client can easily do so, and simplifies this package
- Change signatures of internal routines to be access-to-USART

serial_io-nonblocking.adb
- Internal simplifications reflecting change to access discriminant now designating type USART
- Remove code for Initialize_Hardware and Configure, per removal from package spec
- Use short-circuit control forms in ISR since don't need to check both conditions, should be
a little faster

serial_io-streaming.ads
- Type Serial_Port now designates a USART rather than type Peripheral
- No longer provide call-through routines for Serial_IO packages
Initialize_Hardware and Configure since client can easily do so, and simplifies this package
- Change signatures of internal routines to be access-to-USART

serial_io-streaming.adb
- Internal simplifications reflecting change to access discriminant now designating type USART
- Remove code for Initialize_Hardware and Configure, per removal from package spec

serial_io.ads
- Procedure Initialize_Hardware's formal parameter is no longer an access parameter
- Procedure Configure's first formal parameter is now access to USART

serial_io.adb
- Change to procedures' signatures as per spec
- Procedure Configure now uses distinguished receiver syntax for USART routines, much cleaner

### DIFF
--- a/examples/shared/serial_ports/host_app/README.md
+++ b/examples/shared/serial_ports/host_app/README.md
@@ -1,0 +1,3 @@
+This host application is for use specifically with the streaming Serial_Port demonstration main program running on the target board. This host app is required so that the string bounds are handled correctly.
+
+The other demonstration mains do not use streams. For those, use a host app such as TerraTerm or RealTerm.
--- a/examples/shared/serial_ports/host_app/host.adb
+++ b/examples/shared/serial_ports/host_app/host.adb
@@ -1,6 +1,6 @@
 ------------------------------------------------------------------------------
 --                                                                          --
---                     Copyright (C) 2016, AdaCore                          --
+--                   Copyright (C) 2016-2025, AdaCore                       --
 --                                                                          --
 --  Redistribution and use in source and binary forms, with or without      --
 --  modification, are permitted provided that the following conditions are  --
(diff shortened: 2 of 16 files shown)


Fabien-Chouteau (approved):

pat-rogers:
On 09-Sep-25 11:21 AM, Fabien Chouteau wrote:
> ***@***.**** approved this pull request.
Thanks!

PR #450, opened 2025-12-29
Added support for Bmp280 thermometer-barometer sensor

Max-Ogorodnik2006:


pat-rogers (commented):
Well done.
But the type names with suffix "_Type" are inconsisrtent with the ADL naming scheme. For example, these are well named:
   type Pressure is delta 0.0001 range 300.0 .. 1100.0; -- hPa
   type Temperature is delta 0.01 range -40.0 .. 85.0;  -- °C
(although I think Barometric_Pressure would be more descriptive),
whereas these are not very helpful to the reader:
   type Status_Type is private;
   type Power_Mode_Type is (Sleep, Forced, Normal) with Size => UInt2'Size;
   type Precision_Type is

For example, maybe Power_Mode_Type could be Power_Modes which is what the datasheet uses. Then the formal parameter name could just be Mode, which wouldn't be confusing because the procedure is named Set_Power_Mode.

Even BMP280_Device could be BMP280_Pressure_Sensor since that's the term the datasheet uses, and then the reader would know what the device actually is.

Max-Ogorodnik2006:
@pat-rogers 
Thank you for the suggestions. I've implemented them.

pat-rogers (approved):
Looks really good!

ogorodnik (approved):

PR #452, opened 2026-01-05
Add Flash memory driver for STM32 embedded flash.

reznikmm:


jklmnn:
Hi Max,
can you please rebase to the latest master. This should fix the CI.

reznikmm:
Thanks, forgot about it.

jklmnn:
Looks good to me. Can you please add this to `scripts/build_all_examples.py` so that it is built in the CI?

reznikmm:
@jklmnn I've added the test. @ogorodnik found several issues while testing it on his board. Now they are fixed!

arch/ARM/STM32/drivers/flash/stm32-flash.adb
```diff
@@ -0,0 +1,262 @@
...
+      Size   : Positive;
+   begin
+      if Region = Whole_Flash then
+         raise Program_Error with "Unimplemented";
+      end if;
+
+      case Region.From is
+         when 16#0_0000# .. 16#0_FFFF# =>
+            Size := 16 * 1024;
+            Sector := (Region.From - 16#0_0000#) / Size;
```
jklmnn, line 102: This is `Region.From - 0`, is that intended?

PR #455, opened 2026-03-23
Fix potential display coherency problem when D-cache is enabled

pat-rogers:
Affects: stm32-dma2d_bitmap, framebuffer_ltdc

DMA2D_Fill_Rect was called without Synchronous => True, allowing DMA2D writes to the hidden framebuffer to still be in progress when Internal_Update_Layer flushed and handed the buffer to LTDC.  We need to change the call to Fill_Rect to use Synchronous => True, consistent with Fill.

After the calls to Wait_Transfer in procedure Internal_Update_Layer , we need a call to Clean_DCache covering the hidden buffer, before the call to Set_Frame_Buffer, ensuring CPU-written dirty cache lines are flushed to physical memory before LTDC DMA reads them.

Why Wait_Transfer alone is insufficient: Wait_Transfer waits for any in-progress DMA2D transfer to complete on that buffer, and then calls Clean_Invalidate_DCache. That's correct for a buffer the CPU is about to read (invalidate stale CPU-cache lines so it re-fetches DMA2D-written data). But for a buffer LTDC is about to read, we need Clean_DCache (flush CPU-dirty lines to physical memory without discarding them).  In other words, Clean_Invalidate is correct for the CPU-read case, and Clean only is correct for the LTDC-handoff case. The Clean_Invalidate_DCache inside Wait_Transfer becomes redundant given the subsequent Clean_DCache, but the synchronization barrier it provides is not. We still need to wait for DMA2D to finish before flipping.

### DIFF
--- a/boards/stm32_common/dma2d/stm32-dma2d_bitmap.adb
+++ b/boards/stm32_common/dma2d/stm32-dma2d_bitmap.adb
@@ -1,6 +1,6 @@
 ------------------------------------------------------------------------------
 --                                                                          --
---                     Copyright (C) 2015-2016, AdaCore                     --
+--                     Copyright (C) 2015-2026, AdaCore                     --
 --                                                                          --
 --  Redistribution and use in source and binary forms, with or without      --
 --  modification, are permitted provided that the following conditions are  --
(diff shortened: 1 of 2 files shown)


pat-rogers:
Problem remains under at least one as-yet-undetermined circumstance. Investigating further...

jklmnn (approved):

PR #457, opened 2026-03-24
Fix SDRAM timing setup during initialization

pat-rogers:
stm32-sdram.adb: 
We now compute initialization values from the datasheet constants. Specifically, in procedure Initialize:

- ExitSelfRefreshDelay, SelfRefreshTime, and RowCycleDelay were all derived from a single SDRAM_Min_Delay_In_ns constant, improperly collapsing three distinct timing constraints (TXSR, TRAS, TRC) into one value. 

- SelfRefreshTime was also hardcoded to 4 cycles, valid only at 90 MHz SDCLK.

We replace those with per-parameter ceiling division over named board constants (SDRAM_TXSR_In_Ns, SDRAM_TRAS_In_Ns, SDRAM_TRC_In_Ns) so each constraint is satisfied independently at any clock frequency.

stm32-board.ads:

- add new SDRAM constants for the sake of the SDRAM initialization

### DIFF
--- a/boards/stm32_common/stm32f429disco/stm32-board.ads
+++ b/boards/stm32_common/stm32f429disco/stm32-board.ads
@@ -132,7 +132,12 @@ package STM32.Board is
    SDRAM_Read_Pipe       : constant STM32.FMC.FMC_SDRAM_Read_Pipe_Delay :=
                              STM32.FMC.FMC_ReadPipe_Delay_1;
    SDRAM_Refresh_Cnt     : constant := 1386;
-   SDRAM_Min_Delay_In_ns : constant := 70;
+
+   --  SDRAM AC timings (ns) for the on-board ISSI IS42S16400J
+   --  Values from the ST BSP in stm32f429i_discovery_sdram.c at SDCLK = 90 MHz:
+   SDRAM_TRAS_In_Ns      : constant := 42;  --  TRAS  
+   SDRAM_TRC_In_Ns       : constant := 70;  --  TRC   
+   SDRAM_TXSR_In_Ns      : constant := 70;  --  TXSR
 
    ---------------
    -- SPI5 Pins --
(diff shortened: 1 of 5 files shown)


pat-rogers:
@jklmnn  The change to stm32-board.ads doesn't seem to be taken into account by the CI builder, so as a result the change to stm32-sdram.adb fails.

jklmnn:
> @jklmnn The change to stm32-board.ads doesn't seem to be taken into account by the CI builder, so as a result the change to stm32-sdram.adb fails.

`stm32-sdram.adb` is used for all STM32 boards, so the new values need to be added to all `stm32-board.ads` implementations. The test script for example fails to build the example for the STM32F429.

jklmnn (changes requested):
See https://github.com/AdaCore/Ada_Drivers_Library/pull/457#issuecomment-4327907552.

boards/stm32_common/stm32f429disco/stm32-board.ads
```diff
@@ -132,7 +132,12 @@ package STM32.Board is
    SDRAM_Read_Pipe       : constant STM32.FMC.FMC_SDRAM_Read_Pipe_Delay :=
                              STM32.FMC.FMC_ReadPipe_Delay_1;
    SDRAM_Refresh_Cnt     : constant := 1386;
-   SDRAM_Min_Delay_In_ns : constant := 70;
+
+   --  SDRAM AC timings (ns) for the on-board ISSI IS42S16400J
+   --  Values from the ST BSP in stm32f429i_discovery_sdram.c at SDCLK = 90 MHz:
+   SDRAM_TRAS_In_Ns      : constant := 42;  --  TRAS  
+   SDRAM_TRC_In_Ns       : constant := 70;  --  TRC
```
jklmnn: These two lines have trailing spaces. Removing these should fix the CI.

jklmnn (changes requested):

jklmnn (approved):

PR #461, opened 2026-03-28
Ensure loop does not go forever in RNG

pat-rogers:
the loop that checks for data ready is otherwise potentially infinite. not seen in practice, but no need to allow it.

### DIFF
--- a/arch/ARM/STM32/drivers/stm32-rng-polling.adb
+++ b/arch/ARM/STM32/drivers/stm32-rng-polling.adb
@@ -1,6 +1,6 @@
 ------------------------------------------------------------------------------
 --                                                                          --
---                    Copyright (C) 2015, AdaCore                           --
+--                  Copyright (C) 2015-2026, AdaCore                        --
 --                                                                          --
 --  Redistribution and use in source and binary forms, with or without      --
 --  modification, are permitted provided that the following conditions are  --
@@ -50,9 +50,7 @@ package body STM32.RNG.Polling is
    begin
       Enable_RNG_Clock;
       Enable_RNG;
-
-      --  Discard the first randomly generated number, according to STM32F4
-      --  docs.
+      --  Discard the first generated number, according to STM32F4 docs.
       Discard := Random;
    end Initialize_RNG;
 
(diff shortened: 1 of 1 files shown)


arch/ARM/STM32/drivers/stm32-rng-polling.adb
```diff
@@ -61,10 +59,17 @@ package body STM32.RNG.Polling is
    ------------
 
    function Random return UInt32 is
+      Max_Attempts : constant := 1000;
+      --  The above is arbitrary. It needs to be long enough that normal
+      --  behavior will not be flagged. Longer than that is fine. The point
+      --  is to not hang forever.
```
jklmnn, line 65: According to the documentation the RNG is guaranteed to be filled every 40 RNG clock cycles. Do we have an idea of how many iterations we could do within these 40 clock cycles?

pat-rogers:
On 14-Apr-26 6:06 AM, JK wrote:
> ***@***.**** commented on this pull request.
> 
> ------------------------------------------------------------------------
> 
> In arch/ARM/STM32/drivers/stm32-rng-polling.adb <https://github.com/ 
> AdaCore/Ada_Drivers_Library/pull/461#discussion_r3078978323>:
> 
>> +      Max_Attempts : constant := 1000;
> +      --  The above is arbitrary. It needs to be long enough that normal
> +      --  behavior will not be flagged. Longer than that is fine. The point
> +      --  is to not hang forever.
> 
> According to the documentation the RNG is guaranteed to be filled every 
> 40 RNG clock cycles. Do we have an idea of how many iterations we could 
> do within these 40 clock cycles?

No, I don't. I chose 1000 on the assumption that it should be much more 
than enough.

jklmnn (approved):

jklmnn:
> No, I don't. I chose 1000 on the assumption that it should be much more than enough.

Unfortunately I don't have the facilities to test it but I assume you have tested it on hardware at least once?

pat-rogers:
On 17-Apr-26 5:16 AM, JK wrote:
> *jklmnn* left a comment (AdaCore/Ada_Drivers_Library#461) <https:// 
> github.com/AdaCore/Ada_Drivers_Library/pull/461? 
> email_source=notifications&email_token=ADRX2IKV6MK3QYTIWMAHE334WH77VA5CNFSNUABFM5UWIORPF5TWS5BNNB2WEL2JONZXKZKDN5WW2ZLOOQXTIMRWG4YTINZSGEZ2M4TFMFZW63VGMF2XI2DPOKSWK5TFNZ2LK4DSL5RW63LNMVXHIX3POBSW4X3DNRUWG2Y#issuecomment-4267147213>
> 
>     No, I don't. I chose 1000 on the assumption that it should be much
>     more than enough.
> 
> Unfortunately I don't have the facilities to test it but I assume you 
> have tested it on hardware at least once?

Yes, daily. I use it in an app I'm developing.

PR #456, opened 2026-03-23
replace curly quotes with straight quotes  in comments 

pat-rogers:
GNAT Studio cannot display them

PR #464, opened 2026-04-06
Add ISB to Memory_Barriers for sake of completeness

pat-rogers:
And mark both routines as Inline_Always. Note that is a change for DSB which was just marked Inline, but these don't make sense when not inline.


pat-rogers:
Same problem -- rebasing and pushing did not trigger the checks.

jklmnn:
See https://github.com/AdaCore/Ada_Drivers_Library/pull/456#issuecomment-4243561228.

jklmnn (approved):

PR #465, opened 2026-04-06
Add CRC unit for STM32F7x MCUs

pat-rogers:
The CRC unit on the F7x MCUs (and others) has the additional ability to define the initial value and the polynomial value, but the same core capabilities of the F4 are present. 

* devices/stm32f7x/stm32-device.adb and ads: 
  - Add declaration of CRC unit and procedure to enable the clock
  - Correct comment block header describing the package spec (it had said it was for F4x)

* arch/ARM/STM32/drivers/crc_stm32f7/*
  - Add new source folder and source files for the STM32 CRC unit package spec and body, with the DMA child package.
  - Add readme.txt decribing these new files, explaining that the F7x has additional capabilities but the same core 
    abilities as the F4x. These files are just a copy of the F4 sources, but that way we can enhance them to add the F7x 
    capabilities eventually, without precluding their use with Fx devices in the meantime. Worth exploring use of Ada's 
   static inheritance, using the F4 version as the parent type, but IFF the additional F7 capabilities don't require additional
   record components.

* boards/stm32f746_discovery/stm32f746_discovery_full.gpr:
    - Manually added source new folder for the F7x CRC
    
* scripts/config/devices.py:
  - Added source new folder for the F7x CRC

### DIFF
--- a/boards/stm32f746_discovery/stm32f746_discovery_full.gpr
+++ b/boards/stm32f746_discovery/stm32f746_discovery_full.gpr
@@ -114,6 +114,7 @@ library project STM32F746_Discovery_Full is
     Src_Dirs_Root & "/arch/ARM/STM32/drivers/power_control_stm32f7", -- From MCU definition
     Src_Dirs_Root & "/arch/ARM/STM32/drivers/ltdc/", -- From MCU definition
     Src_Dirs_Root & "/arch/ARM/STM32/drivers/sai/", -- From MCU definition
+    Src_Dirs_Root & "/arch/ARM/STM32/drivers/crc_stm32f7/", -- manually added
     Src_Dirs_Root & "/arch/ARM/STM32/drivers/sd/", -- From MCU definition
     Src_Dirs_Root & "/arch/ARM/STM32/drivers/sd/sdmmc/", -- From MCU definition
     Src_Dirs_Root & "/middleware/src/filesystem", -- From middleware config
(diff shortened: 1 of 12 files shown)


pat-rogers:
For some reason the other files in the commit are not being taken into account by the builder, IIUC.

boards/stm32f746_discovery/stm32f746_discovery_full.gpr
```diff
@@ -114,6 +114,9 @@ library project STM32F746_Discovery_Full is
     Src_Dirs_Root & "/arch/ARM/STM32/drivers/power_control_stm32f7", -- From MCU definition
     Src_Dirs_Root & "/arch/ARM/STM32/drivers/ltdc/", -- From MCU definition
     Src_Dirs_Root & "/arch/ARM/STM32/drivers/sai/", -- From MCU definition
+
+    Src_Dirs_Root & "/arch/ARM/STM32/drivers/crc_stm32f7/", -- manually added but the MCU definition has been updated
+
```
jklmnn: Can you please remove the empty lines for consistency?

jklmnn:
> For some reason the other files in the commit are not being taken into account by the builder, IIUC.

You also have to add the new directory to `boards/stm32f746_discovery/stm32f746_discovery_sfp.gpr`.

jklmnn (changes requested):

boards/stm32f746_discovery/stm32f746_discovery_full.gpr
pat-rogers: Done. I added the gpr files for the F769 boards too.

jklmnn (approved):

Issue #466, opened 2026-04-08
Fix CI


jklmnn:
The CI is currently broken. This needs to be fixed before any PRs can be merged.

PR #467, opened 2026-04-08
Use pragma instead of aspect for Interrupt_Priority on Serial_Port

jklmnn:
The definition of Serial_Port uses the aspect Interrupt_Priority taking a value from the specified discriminants. However the discriminant is not visible in this scope. Use a pragma in the definition of the type instead where the discriminant is visible.

ref #466

### DIFF
--- a/examples/shared/serial_ports/src/serial_io-nonblocking.ads
+++ b/examples/shared/serial_ports/src/serial_io-nonblocking.ads
@@ -77,9 +77,11 @@ private
      (Device       : not null access USART;
       IRQ          : Interrupt_ID;
       IRQ_Priority : Interrupt_Priority)
-   with
-      Interrupt_Priority => IRQ_Priority
    is
+      pragma Interrupt_Priority (IRQ_Priority);
+      --  FIXME: this is supposed to be an aspect, however with FSF GNAT 12
+      --  the compiler does not have IRQ_Priority in scope when an aspect is
+      --  used and rejects it.
 
       procedure Start_Sending (Msg : not null access Message);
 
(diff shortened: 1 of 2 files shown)


pat-rogers:
Hello Johannes,

On 08-Apr-26 9:19 AM, JK wrote:
> The definition of Serial_Port uses the aspect Interrupt_Priority taking 
> a value from the specified discriminants. However the discriminant is 
> not visible in this scope. Use a pragma in the definition of the type 
> instead where the discriminant is visible.
> 
> ref #466 <https://github.com/AdaCore/Ada_Drivers_Library/issues/466>

I believe that's a regression in the compiler, unless there has been a 
decision by the ARG. One of the releases of the compiler rejected it in 
the now-distant past but later compilers accepted it. I have used that 
idiom many times successfully, so unless there has been a recent ruling 
against it, it should still be legit. Has something changed?

Thanks,

Pat

jklmnn:
Ah, thanks for the heads-up. I will see if I can reproduce the internally and open an issue for that, or if it can be fixed with a newer FSF GNAT build. If we can't immediately fix it here, I'd still go for the current workaround (and add a comment) to have at least the CI working again so we have some testing for the other PRs.

pat-rogers:
Agreed. 
I've been using GNAT Pro versions so there could indeed still be a problem with the FSF version, although I would have thought that it would have caught up with the Pro verion by now. 
Anyway, thanks!

zakaria1193-adacore (approved):
LGTM

jklmnn:
@pat-rogers can you rebase your PRs and see if the CI is fixed in them? Thanks!

pat-rogers:
Hello Johannes,
  Note that the use of the aspect compiles successfully with the 
gnat-15.2.0-1 FSF compiler.

Best regards,

Pat

PR #470, opened 2026-05-05
Add new Device_Id package body for F7x, which uses a different address

pat-rogers:
The F7x series supports unique device identifiers but the id value is at a diferent address than the F4x series. The package interface is the same, therefore we create separate directories for the different package bodies containing the different addresses. One directory for the F4x containing the original package body, one for the F7x for the new package body with the new address.

Modified the board-specific gpr files to reference these device-specific directories.

Modified the Python file that generates the board-specific gpr files to include these new directories.


jklmnn (changes requested):
You also have to adjust [boards/stm32_f4ve/stm32_f4ve_full.gpr](https://github.com/AdaCore/Ada_Drivers_Library/blob/master/boards/stm32_f4ve/stm32_f4ve_full.gpr).

pat-rogers:
I added the F4ve gpr files but the builds still fail. Unfortuantely I cannot see the details to know what causes the failures. When I expand them all I can see is "Error" and nothing else.

jklmnn:
> I added the F4ve gpr files but the builds still fail. Unfortuantely I cannot see the details to know what causes the failures. When I expand them all I can see is "Error" and nothing else.

It seems that [boards/stm32f4xx_m/stm32f4xx_m_full.gpr](https://github.com/AdaCore/Ada_Drivers_Library/blob/master/boards/stm32f4xx_m/stm32f4xx_m_full.gpr) is also missing the entry.

I wasn't aware that the log of the CI is not public. If you want to check it locally you can run
```
python3 scripts/build_all_examples.py
```
The CI is also running
```
python3 testsuite/run.py
python3 scripts/check_readme.py
```
but that is usually not where it fails.

pat-rogers:
> I wasn't aware that the log of the CI is not public. 
 >
> If you want to check it locally you can run
> 
> |python3 scripts/build_all_examples.py |

OK, thanks. Will do.

It would be good to make it publicly visible.

jklmnn (approved):

PR #475, opened 2026-05-10
The color named Chart_Reuse -> Chartreuse

pat-rogers:
I was tempted to leave it as-is...  :-)

### DIFF
--- a/hal/src/hal-bitmap.ads
+++ b/hal/src/hal-bitmap.ads
@@ -423,4 +423,7 @@ package HAL.Bitmap is
    White_Smoke         : constant Bitmap_Color := (255, 245, 245, 245);
    White               : constant Bitmap_Color := (255, 255, 255, 255);
 
+   Chart_Reuse         : Bitmap_Color renames Chartreuse;
+   --  Retain the original 'spelling' for backwards compatibility
+
 end HAL.Bitmap;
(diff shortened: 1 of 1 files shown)


jklmnn:
What is the intention for this change (it may be a language thing I'm missing here)? The problem I see is that it effectively changes the API of `HAL.Bitmap`. So every alire crate using this interface needs to be updated if it happens to use that constant.

It would also warrant a new version number for the [alire crate](https://alire.ada.dev/crates/hal.html).

pat-rogers:
> What is the intention for this change (it may be a language thing I'm 
> missing here)? 

"Chart_Reuse" is a misspelling of the color "Chartreuse". I assume it 
was a (natural) language thing initially because it is a French word. 
Maybe the effect of a spelling corrector? I didn't look to see who wrote 
that line though.

> The problem I see is that it effectively changes the API 
> of |HAL.Bitmap|. So every alire crate using this interface needs to be 
> updated if it happens to use that constant.
> 
> It would also warrant a new version number for the alire crate <https:// 
> alire.ada.dev/crates/hal.html>.

Your call. We can close the PR if you think it isn't worth it. I don't 
use that constant myself.

jklmnn:
> "Chart_Reuse" is a misspelling of the color "Chartreuse". I assume it
was a (natural) language thing initially because it is a French word.
Maybe the effect of a spelling corrector? I didn't look to see who wrote
that line though.

Oh, my head thought it was about reusing a chart lol.

> Your call. We can close the PR if you think it isn't worth it. I don't use that constant myself.

Can you add the correct version but keep the incorrect one with a comment that we leave it for backwards compatibility?

jklmnn (approved):

PR #479, opened 2026-06-27
Fix path to svd2ada binary in svd.mk

A-Yerkesh:
I believe svd.mk assumes outdated structure for svd2ada tool. Currently, binary is located in bin/ directory.

Running svd.mk updates many existing driver files. Is this expected or something is wrong?

I had problem running svd.mk, because STM32F7x9.svd in svd2ada/CMSIS-SVD/ST/ was corrupted or outdated. Here is the error i was getting:
```
Fatal error when parsing the SVD file:
STM32F7x9.svd:22934:7: Name differ for closing tag (expecting register, opened line 22567)
```

I fixed it by downloading recent version from [here](https://github.com/cmsis-svd/cmsis-svd-data/). I plan to make a pull request with this update to svd2ada repository soon.

### DIFF
--- a/arch/svd.mk
+++ b/arch/svd.mk
@@ -8,21 +8,21 @@ all: svd
 
 svd:
 	rm -rf $(STM_DIR)/stm32*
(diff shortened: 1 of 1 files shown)


jklmnn (approved):
Thanks for the fix!

